Issue #490, opened 2021-04-01
credential issues with Bigquery


set5think:
First off: excellent project.  Thank you for creating it!

I was trying to set up a Bigquery query engine last night.  I pasted in my json credentials into the field in the UI, and it parsed correctly.  When I went to run a query though, I would receive an error `'ClientOptions' object has no attribute 'credentials_file'`

I was able to resolve this by adding `google-cloud-core==1.3.0` to requirements/base.txt

czgu:
Thanks for reporting the issue! I dig around https://github.com/googleapis/python-bigquery and it seems like we would need to update its version to be at least this 
```
google-cloud-bigquery==1.28.0
```

sadly I don't have a big query account now to test the change. If it works for you, can you submit a PR to update the version or just let me know so I can submit one?

set5think:
hey yea I'll open a PR.  I'm working through the dependency issues right now.  Upgrading bigquery to 1.28 created an incompatibility with google-resumable-media and gcs.   I'll push my changes up shortly, thanks!

PR #494, opened 2021-04-02
fix: update google packages.

set5think:
These updates account for allowing client_options to be passed to the
google auth client, which allows credentials to be provided via json
instead of ENV vars.


set5think:
resolves #490

czgu (approved):

PR #503, opened 2021-04-06
Makefile updates

trevorgrayson:
- yarn install is now idempotent.
- dependent targets updated

consider installing python packages to a dir, you can make that
idempotent too e.g. http://makef.io/python

This is a bit contrived, but I like using this with Makefiles: http://necessaryeval.com/2020/02/27/dsky/


czgu (approved):

czgu:
thanks for the contribution! good idea to make python idempotent too

PR #505, opened 2021-04-09
Modify cellType title

vnnw:
Read cellType title from 'name' field of 'datadoc.yaml' for better UI customizations, such as using alternative terms or translating into other languages.

### DIFF
--- a/querybook/webapp/components/DataDoc/DataDocCellControl.tsx
+++ b/querybook/webapp/components/DataDoc/DataDocCellControl.tsx
@@ -10,7 +10,16 @@ import { SoftButton } from 'ui/Button/Button';
 import { Dropdown } from 'ui/Dropdown/Dropdown';
 import { ListMenu, IListMenuItem } from 'ui/Menu/ListMenu';
 
-const cellTypes = require('config/datadoc.yaml').cell_types;
+const cellTypes: Record<
+    string,
+    {
+        key: string;
+        icon: string;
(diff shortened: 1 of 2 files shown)


querybook/webapp/components/DataDoc/DataDocCellControl.tsx
```diff
@@ -268,7 +268,10 @@ const InsertCellButtons: React.FC<{
             key={cellKey}
             onClick={() => handleInsertcell(cellKey)}
             icon="plus"
-            title={titleize(cellKey)}
+            title={ !cellTypes[cellKey].name
```
czgu, line 271: can you change the iteration to 
```
Object.entries(cellTypes).map(([cellKey, cellType]) => (
```
since cellTypes[cellKey] is needed


and can we change title to `title={cellType.name ?? titleize(cellKey)}`

czgu:
thanks for contributing, btw to fix the lint error you can run prettier before submitting the PR 😃

czgu:
😂  looks like it was a bad idea to not type yaml imports, i guess you can change line 13 to be
```
const cellTypes: Record<
    string,
    {
        key: string;
        icon: string;
        name?: string;
        meta: Record<string, unknown>;
        meta_default: Record<string, unknown>;
    }
> = require('config/datadoc.yaml').cell_types;
```

czgu (approved):

PR #519, opened 2021-04-22
ui: add max width to table view

meowcodes:
looks same in modal & is centered like a datadoc in non-modal view


czgu:
use <Container /> styled component with .DataTableView className?

czgu (approved):

PR #529, opened 2021-04-27
fix: Improve socketio error handling behaivor

czgu:
- Refactored socket io datasource code
- Require socket.io connection to be always authenticated
- Debounce the error message on disconnect event, if they are reconnected with 2 seconds, then do not show the error message to the user

### DIFF
--- a/querybook/webapp/lib/socketio-manager.ts
+++ b/querybook/webapp/lib/socketio-manager.ts
@@ -1,4 +1,4 @@
-import { throttle } from 'lodash';
+import { debounce } from 'lodash';
 import toast from 'react-hot-toast';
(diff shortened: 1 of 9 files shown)


justinmpier (approved):

querybook/webapp/lib/socketio-manager.ts
```diff
@@ -19,9 +19,13 @@ function getSocketFromManager(nameSpace: string) {
     return socketIOManager.socket(nameSpace);
 }
 
-const sendToastForError = throttle((error) => {
-    toast.error(String(error));
-}, 3000);
+const sendToastForError = debounce(
```
meowcodes, line 22: sendErrorToast?

PR #536, opened 2021-05-09
fix: support https elasticsearch host

Aka-shi:
This PR adds `https`  as one of the values for `ELASTICSEARCH_CONNECTION_TYPE` to allow specifying https url for elasticsearch host.

### DIFF
--- a/querybook/server/logic/elasticsearch.py
+++ b/querybook/server/logic/elasticsearch.py
@@ -49,6 +49,8 @@ def get_hosted_es():
 
     if QuerybookSettings.ELASTICSEARCH_CONNECTION_TYPE == "naive":
         hosted_es = Elasticsearch(hosts=[host], port=port,)
+    elif QuerybookSettings.ELASTICSEARCH_CONNECTION_TYPE == "https":
+        hosted_es = Elasticsearch(hosts=[host], port=port, scheme="https",)
     elif QuerybookSettings.ELASTICSEARCH_CONNECTION_TYPE == "aws":
         # TODO: generialize aws region setup
         from boto3 import session as boto_session


querybook/server/logic/elasticsearch.py
```diff
@@ -49,6 +49,8 @@ def get_hosted_es():
 
     if QuerybookSettings.ELASTICSEARCH_CONNECTION_TYPE == "naive":
         hosted_es = Elasticsearch(hosts=[host], port=port,)
+    elif QuerybookSettings.ELASTICSEARCH_CONNECTION_TYPE == "https":
```
czgu, line 52: i checked the documentation and to use https you would need to set `use_ssl=True` instead of scheme='https'? Also for extensibility, maybe we can do this instead?

```
if QuerybookSettings.ELASTICSEARCH_CONNECTION_TYPE == "naive":
    hosted_es = Elasticsearch(hosts=[host], port=port, use_ssl=host.startsWith('https'))
```

Aka-shi: `host.startsWith('https')` wouldn't work here. Because we are splitting the connection string above this on `:`. For example if `ELASTICSEARCH_HOST = https://localhost:1234`, then we would have `host = https` which is wrong.

Aka-shi: >i checked the documentation and to use https you would need to set use_ssl=True instead of scheme='https'

Not sure about this, it worked for us by just specifying `scheme=https` as the endpoint was accessible through https but the elasticsearch cluster didn't have any auth enabled. When we specify `use_ssl=True`, it expects the elasticsearch cluster to have ssl enabled if I am not wrong. Which I think would also need to be supported sometime in future. 

@czgu thoughts?

Issue #565, opened 2021-06-18
Metadata loader should support catalog.schema.table format


anismiles:
Enable metadata loaders to scan and index stores that support 3 levels of hierarchies, for example - the metadata loader for Presto should be able to scan all catalogs, their schemes, and tables using a single loader.

PR #566, opened 2021-06-18
Feat: support catalog and schema in metadata

anismiles:
Fixes: #565

For sources where `schemas` live under a `catalog`, like `Presto`,  catalog and schema can be merged into a single name `catalog.schema` and stored in the database.


czgu (changes requested):
Hi Animesh, thanks for submitting the PR. I am concerned with the approach of having "catalog.schema"  in schema, since catalog would map to a metastore.

anismiles:
When you have tens/hundreds of Catalogs in the cluster, making separate Metastore for each of them would - 
1. be cumbersome to setup
2. not allow cross catalog queries. (since runner is also tied with metastore) 

Animesh

PR #575, opened 2021-06-25
feat: change workflow to publish to dockerhub

czgu:
Instead of publishing to github, publish the prod image to dockerhub


czgu:
Tested publish in my local fork, trying to merge now to see if secrets can be accessed at that level.

meowcodes (approved):

Issue #578, opened 2021-06-29
Necessary environment variable for configuring s3 as result storage


yongchand:
In order to configure S3 as a result storage, **AWS_ACCESS_KEY_ID** and **AWS_SECRET_ACCESS_KEY** are required as environment variable for boto3.
Even though people who are familiar with boto3 may know this fact, I feel like a lot of people will be confused and face "access denied" issue.
Isn't it better to explicitly write this on code or at least on the document?
I am willing to contribute to this :)
Thank you 👍

czgu:
Hi Yongchan, yes that would make sense. When I implemented it internally, the aws_access_key_id and secret key wasn't needed. I was guessing it is because of the iamrole which granted the access.

Would you mind to update the documentation here https://github.com/pinterest/querybook/blob/master/docs_website/docs/integrations/add_result_store.md about the s3 result storage? We can have the format similar to https://github.com/pinterest/querybook/blob/master/docs_website/docs/integrations/add_auth.md. If it is a common problem people face, we can add this to code to make it easier. Thank you!

PR #580, opened 2021-06-30
fix: add explanation about mandatory environment variables

yongchand:
Issue #578 
Thank you :)

### DIFF
--- a/docs_website/docs/integrations/add_result_store.md
+++ b/docs_website/docs/integrations/add_result_store.md
@@ -18,3 +18,13 @@ If you cannot include this result store as part of the open source project, you 
 1. Locate the plugin root directory for your customized Querybook, and find the folder called result_store_plugin.
 2. Add your result store code similiar to what's above, which means making sure it inherits from base_store.py.
 3. Add the new reuslt store along with its representative name in the variable ALL_PLUGIN_RESULT_STORES under result_store_plugin/\_\_init\_\_.py
+
+## Necessary Field for Configuration
+
+Following environment variables are required for configuration
+
+### S3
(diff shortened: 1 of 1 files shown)


docs_website/docs/integrations/add_result_store.md
```diff
@@ -18,3 +18,13 @@ If you cannot include this result store as part of the open source project, you
 1. Locate the plugin root directory for your customized Querybook, and find the folder called result_store_plugin.
 2. Add your result store code similiar to what's above, which means making sure it inherits from base_store.py.
 3. Add the new reuslt store along with its representative name in the variable ALL_PLUGIN_RESULT_STORES under result_store_plugin/\_\_init\_\_.py
+
+## Necessary Field for Configuration
+
+Following environment variables are required for configuration
```
czgu, line 24: can you move this under `### S3`

czgu (commented):
Just to follow up with the issue ticket, do you know if this is required or its needed if you need a separate iamrole? 

btw, I would recommend installing pre commit hooks to prevent lint errors in future, here is the guide https://www.querybook.org/docs/developer_guide/developer_setup#setup-local-devtools

yongchand:
@czgu this will be required since boto3 requires credentials to access to s3. For more info, I think this document helps: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html. If people have default credentials file, I think it might be fine? but I just wanted to make sure aws_access_key_id and aws_secret_access_key are required. And also thanks for the pre-commit instruction!

czgu (approved):

PR #661, opened 2021-09-24
Support signature v4 enabled buckets as result store

Aka-shi:


querybook/server/env.py
```diff
@@ -83,6 +83,8 @@ class QuerybookSettings(object):
     STORE_MAX_UPLOAD_CHUNK_NUM = int(get_env_config("STORE_MAX_UPLOAD_CHUNK_NUM"))
     STORE_MAX_READ_SIZE = int(get_env_config("STORE_MAX_READ_SIZE"))
     STORE_READ_SIZE = int(get_env_config("STORE_READ_SIZE"))
+    STORE_BUCKET_S3V4_ENABLED = get_env_config("STORE_BUCKET_S3V4_ENABLED")
+    STORE_BUCKET_REGION = get_env_config("STORE_BUCKET_REGION")
```
czgu, line 87: I am wondering if it would be better if we make this into a more general setting called `AWS_REGION` which can be  used for elasticsearch as well.
Once you made that change, can you update all `us-east-1` with this config value instead?

querybook/server/env.py
```diff
@@ -83,6 +83,8 @@ class QuerybookSettings(object):
     STORE_MAX_UPLOAD_CHUNK_NUM = int(get_env_config("STORE_MAX_UPLOAD_CHUNK_NUM"))
     STORE_MAX_READ_SIZE = int(get_env_config("STORE_MAX_READ_SIZE"))
     STORE_READ_SIZE = int(get_env_config("STORE_READ_SIZE"))
+    STORE_BUCKET_S3V4_ENABLED = get_env_config("STORE_BUCKET_S3V4_ENABLED")
```
czgu, line 86: since STORE_... are kind of shared between GCP and AWS, can you prefix the setting with something like S3? It can be `S3_BUCKET_V3_ENABLED`.

Also since the value is boolean, lets do
```
    STORE_BUCKET_S3V4_ENABLED = get_env_config("STORE_BUCKET_S3V4_ENABLED") == "true"
```
so in the code we can use it as boolean directly

czgu (commented):
Can you add a bit of documentation to `add_result_store.md` to cover the usage? Just some quick notes are good enough 👍

querybook/server/env.py
Aka-shi: Made the changes. Please review.

querybook/server/env.py
Aka-shi: Done.

querybook/config/querybook_default_config.yaml
```diff
@@ -58,6 +58,8 @@ STORE_MIN_UPLOAD_CHUNK_SIZE: 10485760
 STORE_MAX_UPLOAD_CHUNK_NUM: 10000
 STORE_MAX_READ_SIZE: 131072
 STORE_READ_SIZE: 5242880
+STORE_BUCKET_S3V4_ENABLED: false
```
czgu, line 61: can you update these to 
S3_BUCKET_S3V4_ENABLED
AWS_REGION

czgu (commented):
Don't forget to update the documentation :)

czgu (approved):

PR #662, opened 2021-09-27
configurable proxy_user for query executors

Aka-shi:
In our data stack, we use the user's email_id as the user identifier for our data access policies. Querybook by default sets `user.username` as the proxy_user which isn't desirable for everyone. 

This PR makes the proxy_user configurable with possible values of `username` and `email`. It defaults to `username` to avoid breaking current setups.

### DIFF
--- a/querybook/config/querybook_default_config.yaml
+++ b/querybook/config/querybook_default_config.yaml
@@ -44,6 +44,10 @@ OAUTH_AUTHORIZATION_URL: ~
 OAUTH_TOKEN_URL: ~
 OAUTH_USER_PROFILE: ~
 
+# User identifier which will be passed to query engines.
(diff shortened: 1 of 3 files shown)


czgu (commented):
Hmm I think this might not be needed if you want to pass email instead. You can do this instead?

```python
from logic.user import get_user_by_name

class CustomPrestoClient(PrestoClient):
    def __init__(
        self,
        connection_string,
        proxy_user=None, 
        *args,
        **kwargs
    ):
        proxy_user_email = get_user_by_name(proxy_user).email
        super(PrestoClient, self).__init__(
            connection_string, proxy_user=proxy_user_email, *args, **kwargs
        )

```

Aka-shi:
Closing in favor of #673

PR #667, opened 2021-09-28
fix: Improve latest_partition template function

### LINKED ISSUE #650: Minor: Add some additional features to get_latest_partition
1. Right now get_latest_partition cannot be recursively rendered in a variable, we would need to supply a custom jinjaenv to _flatten_variable in order to do so. Let's also add this as a unit test to prevent future cases
2. Update the parameters of get_latest_partition, if the partition is like 'dt=2020-01-01' then the second parameter is not required, however if there are more than 1 parameter, then the second parameter is required (we can throw an error in this case). For example
```
partitions = ['dt=1', 'dt=2', 'dt=3']
get_latest_partition('schema.table') # returns 3

partitions = ['dt=1/a=2', 'dt=2/a=2', 'dt=3/a=3']
get_latest_partition('schema.table') # throws exception
```

3. Additionally, there should be a third optional parameter called `conditions` where you can pass in filters of shape Dict[str, str] such that it would filter the partitions. For example
```
# earliest to latest
partitions = [
    'a=foo/b=1',
    'a=bar/b=1',
    'a=baz/b=2'
]
get_latest_partition('schema.table', 'a', { 'b': 1 }) # returns 'bar' instead of 'baz'
```

kgopal492:
closes #650

Minor improvements to `latest_partition` template function

* Make partition key optional when table has only one partition column
* Allow passing conditions to filter latest partition
* Allow added `latest_partition` function to a recursive template variable

### DIFF
--- a/querybook/server/clients/glue_client.py
+++ b/querybook/server/clients/glue_client.py
@@ -138,8 +138,7 @@ class GlueDataCatalogClient:
         result = []
 
         if conditions:
-            for condition in conditions.items():
-                (condition_key, condition_value) = condition
+            for condition_key, condition_value in conditions.items():
                 condition_key_index = partition_key_names.index(condition_key)
                 partition_values = list(
                     filter(
--- a/querybook/server/clients/glue_client.py
+++ b/querybook/server/clients/glue_client.py
@@ -140,13 +140,11 @@ class GlueDataCatalogClient:
         if conditions:
             for condition_key, condition_value in conditions.items():
                 condition_key_index = partition_key_names.index(condition_key)
-                partition_values = list(
-                    filter(
-                        lambda partition_value: partition_value[condition_key_index]
-                        == condition_value,
-                        partition_values,
-                    )
-                )
+                partition_values = [
+                    partition_value
+                    for partition_value in partition_values
+                    if partition_value[condition_key_index] == condition_value
+                ]
 
         for partition_value in partition_values:
             result.append(
--- a/querybook/server/clients/glue_client.py
+++ b/querybook/server/clients/glue_client.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Dict, List
 
 import boto3
 from env import QuerybookSettings
@@ -93,7 +93,7 @@ class GlueDataCatalogClient:
         Gets partition information for db_name.tb_name from the Glue Data Catalog
 
         :param db_name: The name of the database
-        :param tb_name: Then name of the table
+        :param tb_name: The name of the table
         :return: The Glue partition objects of db_name.tb_name
         """
         _LOG.info(f"Get Glue partitions for ${db_name}.${tb_name}")
@@ -111,13 +111,16 @@ class GlueDataCatalogClient:
 
         return result
 
-    def get_hms_style_partitions(self, db_name, tb_name) -> List[str]:
+    def get_hms_style_partitions(
+        self, db_name, tb_name, conditions: Dict[str, str] = None
+    ) -> List[str]:
         """
         Gets partitiion information for db_name.tb_name from Glue Data Catalog and converts into a
         Hive Metastore style representation
 
         :param db_name: The name of the database
         :param tb_name: The name of the table
+        :param conditions: Filter conditions for the partition in the format { dt: '2016-03-14'}
         :return: The partitions of db_name.tb_name in the format ['dt=2016-03-14/hr=00', 'dt=2016-03-14/hr=01', ...]
         """
         _LOG.info(f"Get hms style partitions for ${db_name}.${tb_name}")
@@ -133,6 +136,18 @@ class GlueDataCatalogClient:
         partition_values = [partition.get("Values") for partition in partition_list]
 
         result = []
+
+        if conditions:
+            for condition in conditions.items():
+                (condition_key, condition_value) = condition
+                condition_key_index = partition_key_names.index(condition_key)
+                partition_values = list(
+                    filter(
+                        lambda partition_value: partition_value[condition_key_index]
+                        == condition_value,
+                        partition_values,
+                    )
+                )
 
         for partition_value in partition_values:
             result.append(
--- a/querybook/server/clients/hms_client.py
+++ b/querybook/server/clients/hms_client.py
@@ -221,9 +221,7 @@ class HiveMetastoreClient:
 
         Returns:  List[str] The partitions keys of the table
         """
-        table = self._perform_read_op(
-            lambda: self._read_client.get_table(db_name, tb_name)
-        )
+        table = self.get_table(db_name, tb_name)
         partition_keys = [key.name for key in table.partitionKeys or []]
         return partition_keys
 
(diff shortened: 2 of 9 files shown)


querybook/server/lib/query_analysis/templating.py
```diff
@@ -149,9 +164,12 @@ def get_templated_variables_in_string(s: str, jinja_env=None) -> Set[str]:
     return filtered_variables
 
 
-def verify_all_variables_are_defined(variables_required, variables_provided):
+def verify_all_variables_are_defined(variables_required, variables_provided, jinja_env):
+    global_variables = get_global_variables(jinja_env)
```
czgu, line 168: jinja_global_variables

querybook/server/lib/query_analysis/templating.py
```diff
@@ -168,28 +186,38 @@ def _flatten_variable(
...
     flattened_variables: Dict[str, str],
+    jinja_env,
 ):
     """ Helper function for flatten_recursive_variables.
         Recursively resolve each variable definition
     """
+    global_variables = get_global_variables(jinja_env)
     var_deps = variables_dag[var_name]
-    for dep_var_name in var_deps:
+    filtered_var_deps = list(filter(lambda var: var not in global_variables, var_deps))
```
czgu, line 196: wait this shouldn't have happened since global variables are skipped in the dag?

querybook/server/lib/query_analysis/templating.py
```diff
@@ -71,6 +71,10 @@ def get_default_variables():
     }
 
 
+def get_global_variables(jinja_env):
```
czgu, line 74: just wondering, what is the shape returned, is it List[str]?

querybook/server/lib/query_analysis/templating.py
```diff
@@ -205,18 +233,22 @@ def flatten_recursive_variables(
...
     variables_dag = {}
 
+    global_variables = get_global_variables(jinja_env)
+
     for key, value in raw_variables.items():
         if not value:
             value = ""
 
-        variables_in_value = get_templated_variables_in_string(value, jinja_env)
+        variables_in_value = get_templated_variables_in_string(value)
```
czgu, line 242: can you filter out the global functions in variables_in_value

querybook/server/lib/query_analysis/templating.py
kgopal492: Yup - a List[str]. Added a docstring to clarify

querybook/server/lib/query_analysis/templating.py
kgopal492: I changed variables_in_value to return global functions as well because templated variables with global functions in the value need to be flattened (otherwise the global function will be rendered the the template as the string of the function call - aka the string `'latest_partition(“table”, “dt”)'` will show up in the final rendered query)

querybook/server/lib/query_analysis/templating.py
kgopal492: As explained below, I left the global function variables in the values so that they will be flattened. (But the global variables are not keys in the dag - since the keys are determined from the raw variables.)

querybook/server/clients/hms_client.py
```diff
@@ -210,18 +210,47 @@ def get_table(self, db_name, tb_name):
...
 
         """
         _LOG.info("Get partitions of %s.%s", db_name, tb_name)
+        if filter:
+            partitions = self._perform_read_op(
+                lambda: self._read_client.get_partitions_by_filter(
+                    db_name, tb_name, filter, -1
+                )
+            )
+            partition_keys = self.get_table_partition_keys(db_name, tb_name)
```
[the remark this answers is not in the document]
kgopal492: [Here](https://github.com/gglanzani/hmsclient/blob/master/hmsclient/genthrift/hive_metastore/ttypes.py#L2982) is where it’s defined. I couldn’t find the specific partition keys in this object.

querybook/server/lib/query_analysis/templating.py
czgu: okay that makes sense!

querybook/tests/test_lib/test_query_analysis/test_templating.py
```diff
@@ -108,17 +110,21 @@ def test_nested(self):
 
 
 class FlattenRecursiveVariablesTestCase(TemplatingTestCase):
+    def setUp(self):
+        self.jinja_env = SandboxedEnvironment()
```
czgu, line 114: can you add a test case here to ensure it can ignore global functions?

querybook/server/lib/query_analysis/templating.py
```diff
@@ -262,10 +316,10 @@ def render_templated_query(
     jinja_env = get_templated_query_env(engine_id)
     try:
         escaped_query = _escape_sql_comments(query)
-        variables_in_query = get_templated_variables_in_string(escaped_query, jinja_env)
+        variables_in_query = get_templated_variables_in_string(escaped_query)
 
         all_variables = get_templated_query_variables(variables, jinja_env)
-        verify_all_variables_are_defined(variables_in_query, all_variables)
+        verify_all_variables_are_defined(variables_in_query, all_variables, jinja_env)
```
czgu, line 322: if we keep variables_in_query = get_templated_variables_in_string(escaped_query, jinja_env) then verify_all_variables_are_defined won't need to check for global funcs

querybook/server/lib/query_analysis/templating.py
czgu: actually I don't think we need get_jinja_global_variables at all since `get_templated_variables_in_string(value, jinja_env)` would filter out the globals

querybook/tests/test_lib/test_query_analysis/test_templating.py
```diff
@@ -353,6 +362,9 @@ def test_invalid_table_name(self):
         )
 
     def test_invalid_partition_name(self):
+        self.metastore_loader_mock.get_latest_partition.return_value = (
```
czgu, line 365: can you also add a test to see if we can define latest_partition in a variable?

querybook/server/lib/query_analysis/templating.py
```diff
@@ -168,28 +206,38 @@ def _flatten_variable(
...
     flattened_variables: Dict[str, str],
+    jinja_env,
 ):
     """ Helper function for flatten_recursive_variables.
         Recursively resolve each variable definition
     """
+    jinja_global_variables = get_jinja_global_variables(jinja_env)
     var_deps = variables_dag[var_name]
-    for dep_var_name in var_deps:
+    filtered_var_deps = [var for var in var_deps if var not in jinja_global_variables]
```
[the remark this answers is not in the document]
kgopal492: removing that altogether since we're not filtering anymore

querybook/server/clients/glue_client.py
```diff
@@ -133,6 +136,18 @@ def get_hms_style_partitions(self, db_name, tb_name) -> List[str]:
 
         result = []
 
+        if conditions:
+            for condition in conditions.items():
+                (condition_key, condition_value) = condition
+                condition_key_index = partition_key_names.index(condition_key)
+                partition_values = list(
```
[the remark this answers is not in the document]
kgopal492: fixed

czgu (approved):

czgu (approved):

PR #672, opened 2021-10-02
fix: Change docker image and update README

yongchand:
Sorry, I should have fixed this earlier.
Since Querybook has its image on Docker Hub now, I am replacing the image to use Docker Hub and update README.


czgu (approved):

czgu:
Thank you!

PR #673, opened 2021-10-04
Allow users to specify proxy_user_indetifier for Presto Engine. 

### LINKED ISSUE #666: Allow different kind of proxy username forwarding behavior for Presto
story: Some query engines require forwarding user email instead of user name, we want to have a generic way that's easy to configure in admin UI.

requirements:
- Add a new executor param called `Proxy user forward json path` (or some better name). This field should be an optional string field.
- Add helper message to explain what this thing does
- If empty, then username is forwarded, otherwise, we will fetch the user from User table, convert it to a dict, and then apply a simple json path to it.
- Have this function as a separate util function so we can reuse it for other applicable query engines

For example, if we want to forward the user email, it should be something like 
image

Aka-shi:
Resolves #666

### DIFF
--- a/querybook/server/lib/query_executor/executor_factory.py
+++ b/querybook/server/lib/query_executor/executor_factory.py
@@ -33,16 +33,21 @@ def _get_executor_params_and_engine(query_execution_id, celery_task, session=Non
     if engine.deleted_at is not None:
         raise ArchivedQueryEngine("This query engine is disabled.")
 
+    proxy_user = user.username
+    executor_params = engine.get_engine_params()
+    if executor_params.get("proxy_user_id", "") != "":
+        try:
+            proxy_user = user.to_dict()[executor_params["proxy_user_id"]]
+        except KeyError as e:
+            raise e
+
     return (
         {
             "query_execution_id": query_execution_id,
             "celery_task": celery_task,
             "query": query,
             "statement_ranges": statement_ranges,
-            "client_setting": {
-                **engine.get_engine_params(),
-                "proxy_user": user.username,
-            },
+            "client_setting": {**executor_params, "proxy_user": proxy_user,},
         },
         engine,
     )
--- a/querybook/server/lib/query_executor/executor_template/templates.py
+++ b/querybook/server/lib/query_executor/executor_template/templates.py
@@ -34,6 +34,13 @@ presto_executor_template = StructFormField(
     username=FormField(regex="\\w+"),
     password=FormField(hidden=True),
     impersonate=FormField(field_type=FormFieldType.Boolean),
+    proxy_user_id=FormField(
+        field_type=FormFieldType.String,
+        helper="""
+<p>User field used as proxy_user. proxy_user will be forwaded to Presto as the session user.</p>
+<p>Defaults to username. Possible values are username, email, fullname </p>
+<p>See [here](https://prestodb.github.io/docs/current/installation/jdbc.html) for more details.</p>""",
+    ),
 )
 
 sqlalchemy_template = StructFormField(
--- a/querybook/server/lib/utils/execute_query.py
+++ b/querybook/server/lib/utils/execute_query.py
@@ -29,13 +29,19 @@ class ExecuteQuery(object):
             Any[][]: Returns the result if sync, otherwise None
         """
         engine = get_query_engine_by_id(engine_id, session=session)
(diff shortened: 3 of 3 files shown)


querybook/server/lib/query_executor/executor_factory.py
```diff
@@ -33,16 +33,21 @@ def _get_executor_params_and_engine(query_execution_id, celery_task, session=Non
     if engine.deleted_at is not None:
         raise ArchivedQueryEngine("This query engine is disabled.")
 
+    proxy_user = user.username
```
czgu, line 36: can you extract this to a separate function with user as input

querybook/server/lib/query_executor/executor_factory.py
```diff
@@ -33,16 +33,21 @@ def _get_executor_params_and_engine(query_execution_id, celery_task, session=Non
...
+    executor_params = engine.get_engine_params()
+    if "proxy_user_id" in executor_params.keys():
+        try:
+            proxy_user = user.to_dict()[executor_params["proxy_user_id"]]
+        except KeyError:
```
czgu, line 41: just throw on KeyError, otherwise this is going to be hard to debug

querybook/server/lib/query_executor/executor_factory.py
```diff
@@ -33,16 +33,21 @@ def _get_executor_params_and_engine(query_execution_id, celery_task, session=Non
     if engine.deleted_at is not None:
         raise ArchivedQueryEngine("This query engine is disabled.")
 
+    proxy_user = user.username
+    executor_params = engine.get_engine_params()
+    if "proxy_user_id" in executor_params.keys():
+        try:
+            proxy_user = user.to_dict()[executor_params["proxy_user_id"]]
```
czgu, line 40: you probably want some default value for executor_params["proxy_user_id"] if it is an empty string, say username

querybook/server/lib/query_executor/executor_factory.py
```diff
@@ -33,16 +33,21 @@ def _get_executor_params_and_engine(query_execution_id, celery_task, session=Non
...
         {
             "query_execution_id": query_execution_id,
             "celery_task": celery_task,
             "query": query,
             "statement_ranges": statement_ranges,
-            "client_setting": {
-                **engine.get_engine_params(),
-                "proxy_user": user.username,
-            },
+            "client_setting": {**engine.get_engine_params(), "proxy_user": proxy_user,},
```
czgu, line 50: **executor_params

querybook/server/lib/query_executor/executor_template/templates.py
```diff
@@ -34,6 +34,13 @@
     username=FormField(regex="\\w+"),
     password=FormField(hidden=True),
     impersonate=FormField(field_type=FormFieldType.Boolean),
+    proxy_user_identifier=FormField(
```
czgu, line 37: will proxy_user_id or proxy_user_identifier be used as the param?

czgu (changes requested):

querybook/server/lib/query_executor/executor_factory.py
Aka-shi: Made changes to raise the same error when this occurs.

querybook/server/lib/query_executor/executor_factory.py
Aka-shi: Added in the above condition that executor_params["proxy_user_id"] shouldn't be null. If null, proxy_user would use the default user.username.

querybook/server/lib/query_executor/executor_template/templates.py
Aka-shi: My Bad. It would be `prox_user_id`. Made the change.

querybook/server/lib/query_executor/executor_factory.py
Aka-shi: Made the change.

querybook/server/lib/query_executor/executor_factory.py
czgu: i think we can simplify it to `if executor_params.get('proxy_user_id', '') != ''`

czgu (approved):

czgu (approved):

Issue #509, opened 2021-04-14
Add plugin for Trino


bitsondatadev:
For those that don't know [there was a split in the Presto community 2 years ago, and the Presto creators are now working on a fork of Presto called Trino (formerly known as PrestoSQL)](https://trino.io/blog/2020/12/27/announcing-trino.html). We want to also include Trino in the list. 

There is not much change needed from the [existing Presto executor](https://github.com/pinterest/querybook/blob/master/querybook/server/lib/query_executor/clients/presto.py) from what I can tell, you would simply need to use the correct Trino Python client. There was a recent addition to create the [Trino sqlalchemy project for superset](https://pypi.org/project/sqlalchemy-trino/). This project comes with a warning that it has not been tested thoroughly yet but could likely be used here as well.

bitsondatadev:
I'm going to post this on the Trino page to see if anyone is interested in picking this up on our end.

czgu:
Hi Brian, for the long-term support of Trino, we could use pyhive https://github.com/dropbox/PyHive for the Querybook integration. That way we can support both query engines with the same integrations. However, it seems that they haven't updated Pypi yet, so I can update the docs to recommend users to use the sqlalchemy-trino package for now

bitsondatadev:
We don't officially support PyHive and there is limited compatibility. I would like to keep this open as this will keep support aválenle from the Trino team.

mosabua:
PyHive seems like a bad idea to use since it basically wraps Hive QL. While this is similar to SQL it is NOT really compatible so it will only sort of work at best. I strongly suggest to use some library that is actually suitable for SQL. 

In the case of Python that would the be the standard implementation of the Python dbapi for Trino, which is the trino-python-client. And it is already used by SQLAlchemy as well. If you have any other driver support that uses either of these the Trino support could be very easy to implement.

bitsondatadev:
@czgu thanks for reopening this. I posted on our site and we have a couple people that are interested in working on this.

czgu:
Hi Brian and Manfred, I checked the trino-python-client and it does support cancel query/getting progress which is everything we need for Querybook integration. If you have any questions regarding the query engine implement feel free to reach out to me here on our slack.

bitsondatadev:
@czgu let me follow up with our community to see if anyone knows of a way to do this or if we can get this implemented to support this.

bitsondatadev:
Will do! Thank you @czgu!

shivammmmm:
[...]

PR #681, opened 2021-10-25
Feat : Trino plugin Integration - Querybook

shivammmmm:
Integrating an official Trino Plugin for query book backed-up with native Trino-python-client.
Issue Link : **https://github.com/pinterest/querybook/issues/509**

### DIFF
--- a/querybook/server/lib/query_executor/clients/trino.py
+++ b/querybook/server/lib/query_executor/clients/trino.py
@@ -1,0 +1,92 @@
+import trino
+from lib.query_executor.base_client import ClientBaseClass, CursorBaseClass
+from lib.query_executor.connection_string.trino import get_trino_connection_conf
+
+
+class TrinoClient(ClientBaseClass):
+    def __init__(
+        self, connection_string, username=None, proxy_user=None, *args, **kwargs
+    ):
+        trino_conf = get_trino_connection_conf(connection_string)
+
+        host = trino_conf.host
+        port = 8080 if not trino_conf.port else trino_conf.port
+
+        connection = trino.dbapi.connect(
+            host=host,
+            port=port,
+            catalog=trino_conf.catalog,
+            schema=trino_conf.schema,
+            user=proxy_user or username,
+            http_scheme=trino_conf.protocol,
+        )
+        self._connection = connection
+        super(TrinoClient, self).__init__()
+
+    def cursor(self):
+        return TrinoCursor(cursor=self._connection.cursor())
+
+
+class TrinoCursor(CursorBaseClass):
+    def __init__(self, cursor):
+        self._cursor = cursor
+        self.rows = []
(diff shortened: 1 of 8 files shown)


querybook/server/lib/query_executor/clients/trino.py
```diff
@@ -0,0 +1,97 @@
...
+        port = 8080 if not trino_conf.port else trino_conf.port
+
+        connection = trino.dbapi.connect(
+            host=host,
+            port=port,
+            catalog=trino_conf.catalog,
+            schema=trino_conf.schema,
+            user=proxy_user or username,
+            http_scheme=trino_conf.protocol,
+            )
```
czgu, line 28: this bracket seems misaligned

querybook/server/lib/query_executor/clients/trino.py
```diff
@@ -0,0 +1,97 @@
+import trino 
+from lib.utils.utils import HTTPBasicAndProxyAuth
```
czgu, line 2: import is not used, can you remove?

querybook/server/lib/query_executor/clients/trino.py
```diff
@@ -0,0 +1,97 @@
...
+    def tracking_url(self):
+        return self._tracking_url
+
+    @property
+    def percent_complete(self):
+        return self._percent_complete
+
+    def _update_tracking_url(self, poll_result):
+        if self._tracking_url is None:
+            self._tracking_url = f"{self._cursor._request._http_scheme}://{self._cursor._request._host}:{self._cursor._request._port}/ui/plan.html?{poll_result['queryId']}"
```
czgu, line 94: turn self._cursor._request into a local variable to make the format string easier to read?

requirements/base.txt
```diff
@@ -68,6 +68,9 @@ boto3==1.9.201
 boto==2.49.0
 moto==2.2.4
 
+# Trino Plugin
+trino==0.305.0
```
czgu, line 72: can you remove Trino as part of the default installation?
The reason is that this won't be a scalable approach to add more query engines.

I know there is no example code for such a use case yet. If you can wait, I will put a PR later this week to remove druid and snowflake dependencies from the repo so you can do something similar.

czgu (commented):
PR looks good to me! Just some minor linting fix thats needed. 

Besides the requirements/base.txt issue, can you also update the documentation of https://www.querybook.org/docs/setup_guide/connect_to_query_engines#overview to have Trino as Tier 2 (we can move to Tier 1 later if it works well) and change the package to trino instead of sqlalchemy-trino?

czgu:
Actually, nvm about the requirements/base.txt change. Just note that it will be removed in requirements/base.txt in 3.0

czgu (approved):

Issue #683, opened 2021-10-26
ElasticSearch 7 support


froukees:
### Story
As new ElasticSearch versions are approaching, ES 6 is starting to be outdated for some company environments.

(pre-discussed on [Slack](https://querybook.slack.com/archives/CHCNR2Y5B/p1633617161027500))

___
#### Suggested solution:

- Use [index_per_document_type](https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html#_index_per_document_type) as it'll keep the same structure of [elasticsearch.yaml](https://github.com/pinterest/querybook/blob/master/querybook/config/elasticsearch.yaml), just with removed types in mappings.
- Adjust [server/logic/elasticsearch.py](https://github.com/pinterest/querybook/blob/cab17ccdc7f296be0dac3c44d4689f49638a587f/querybook/server/logic/elasticsearch.py), [server/datasources/search.py](https://github.com/pinterest/querybook/blob/c81d55eab5047c35f4fbeec9cd289d41a2916468/querybook/server/datasources/search.py) - stop working with `doc_type`(s)
- Make ES7 default for `docker-compose`

___
#### OPEN QUESTION
Do drop-in replacement of ES6 configuration (making ES7 only option), or create another [elasticsearch.yaml](https://github.com/pinterest/querybook/blob/master/querybook/config/elasticsearch.yaml), for example `elasticsearch7.yaml` for ES7, and make ES version configurable? The second option is more flexible, but requires some more handling in [server/logic/elasticsearch.py](https://github.com/pinterest/querybook/blob/cab17ccdc7f296be0dac3c44d4689f49638a587f/querybook/server/logic/elasticsearch.py), [server/datasources/search.py](https://github.com/pinterest/querybook/blob/c81d55eab5047c35f4fbeec9cd289d41a2916468/querybook/server/datasources/search.py).

PR #685, opened 2021-10-27
major: migration to es7

froukees:
Solving #683 

This pull requests contains:
- Changes in ElasticSearch document mappings and calls to ES client
- Changes in `docker-compose`, `k8s`, `helm` ElasticSearch deployments
  - As every deployment uses only one ES server (it doesn't actually create ES cluster there)`discovery.type=single-node` was used. This is necessary in ES7, otherwise it leads to verification of production setup ([bootstrap checks](https://www.elastic.co/guide/en/elasticsearch/reference/current/bootstrap-checks.html)) and `the default discovery settings are unsuitable for production use; at least one of [discovery.seed_hosts, discovery.seed_providers, cluster.initial_master_nodes] must be configured` error. But these options are good only with proper cluster.
  - For `helm` deployment, I've taken the setting of ES environmental variables from `elasticsearch-deployment.yaml` template to `values.yaml` completely, so it's more on the eyes.
- Increased version of _base requirement_ `elasticsearch` library

The most recent version of ElasticSearch 7 (7.15.2) was chosen as the default version, but according to my local tests any ES7 minor/patch version should work.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -118,10 +118,11 @@ services:
             timeout: 30s
             retries: 3
     elasticsearch:
-        image: docker.elastic.co/elasticsearch/elasticsearch:6.6.1
+        image: docker.elastic.co/elasticsearch/elasticsearch:7.15.1
         environment:
             cluster.name: docker-cluster
             bootstrap.memory_lock: 'true'
+            discovery.type: single-node
             ES_JAVA_OPTS: -Xms750m -Xmx750m
         ulimits:
             memlock:
--- a/docs_website/docs/changelog/breaking_change.md
+++ b/docs_website/docs/changelog/breaking_change.md
@@ -7,6 +7,42 @@ slug: /changelog
 
 Here are the list of breaking changes that you should be aware of when updating Querybook:
 
+## v3.0.0
+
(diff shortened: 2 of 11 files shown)


czgu:
@froukees Hi Tomas, can you change the elasticsearch-py version to `elasticsearch==7.13.4`.
Looks like elasticsearch-py throws error after 7.14 since they disabled aws hosted es from working.

Relevant issue: https://github.com/elastic/elasticsearch-py/issues/1667

froukees:
> @froukees Hi Tomas, can you change the elasticsearch-py version to `elasticsearch==7.13.4`. Looks like elasticsearch-py throws error after 7.14 since they disabled aws hosted es from working.
> 
> Relevant issue: [elastic/elasticsearch-py#1667](https://github.com/elastic/elasticsearch-py/issues/1667)

Good to know, I didn't know about this. So I've added a commit with decreasing the version of the Python library (keeping ES Docker image version as before in docker compose etc). Tried local deployment using `make` and looks ok.

czgu (approved):

czgu:
Great work! Thanks for adding es7 migration!

Issue #690, opened 2021-11-05
Hive Metastore Loader - Load only specific schemas, configure partition loading


froukees:
## Story
Loading metadata from Hive Metastore using `HMSMetastoreLoader` can take some time with many objects in the given Metastore. Long processing time leads to setting longer scheduling intervals and to longer time between creation of an object in Hive and its propagation to Querybook WebUI.

But sometimes, not everything in Metastore has to be actually loaded for the users, sometimes it might be actually required to not to provide some schemas in the WebUI to the users through an Environment/Query Engine/Metastore.

Therefore, I'd suggest an update of [hive_metastore_loader.py](https://github.com/pinterest/querybook/blob/master/querybook/server/lib/metastore/loaders/hive_metastore_loader.py) containing:
1. possibility to include/exclude schemas:
    1. option: only include option - downloading all by default, but if some schemas are specified, downloading only these 
    2. option: only include option - downloading nothing by default
    2. option: only exclude option - downloading all except these schemas
2. (optional) possibility to turn off partition analysis:
    - This can speed up loading from Hive Metastore in some specific cases, when partition information is not needed
    - This would be probably global for all schemas, because in per schema configuration it'd make sense only in combination with _include option - downloading nothing by default_

What do you think about it? Does it make sense?

**Impact**: This change might cause some issues with loading existing `HMSMetastoreLoader` in the WebUI when the `metastore_dict` changes structure.

A variant with the option where schemas are included could look like this:
![image](https://user-images.githubusercontent.com/22563157/140538225-0ed3b600-1fb6-4485-806a-173b06cf9825.png)
![image](https://user-images.githubusercontent.com/22563157/140535547-32cc32f6-d719-4800-9ea2-bee54a4b42a7.png)

czgu:
hey Tomas for 1, you can use the ACL denylist/allowlist to prevent the metastore loader from loading specific tables or schemas.
For 2, it sounds good to add a flag in the config to prevent partition loading 👍

froukees:
1. Ah, good to know 🙂 
2. Thanks, will do 👍

PR #694, opened 2021-11-12
minor: load_partitions for hive metastore loaders

froukees:
Solving #690

## Description
This pull request adds `Load_partitions` boolean switch to the Metastore creation form for `HMSMetastoreLoader` and `HMSThriftMetastoreLoader`.

## Solution
I've noticed that this switch is already available in `GlueDataCatalogLoader`, so I've decided to extract the form field to a new module and reuse it from there in all three loaders.

## Backwards compatibility
- If I just tried to change the form and opened a Metastore page that uses `HMSMetastoreLoader`, the front-end returned following error: `[Immer] Immer only supports setting array indices and the 'length' property` caused by an array of URLs still present in the DB for the loader in `metastore_params` field.
  - Therefore, I've included transformation for `HMSMetastoreLoader` loader that changes `metastore_params` into proper dictionary in data migrations.

- If there is an existing metastore that uses `HMSMetastoreLoader` or `HMSThriftMetastoreLoader`, it has a behavior of loading all partitions right now. After applying changes from this pull requests, `Load_partitions` would be turned off by default.
  - Therefore, I've included logic that adds `"load_partitions": True` in `metastore_params` for these two loaders in data migrations.

Another solution would be to create completely new loader classes (`HMSMetastoreLoaderV2`, `HMSThriftMetastoreLoaderV2`) that would not infer with the current DB data. But this approach could lead to big number of very similar loaders after some time, so I wasn't sure about it.

### DIFF
--- a/querybook/server/lib/metastore/loaders/form_fileds.py
+++ b/querybook/server/lib/metastore/loaders/form_fileds.py
@@ -1,0 +1,10 @@
+from lib.form import FormField, FormFieldType
+
+
+load_partitions = FormField(
+    required=False,
+    field_type=FormFieldType.Boolean,
+    helper="""In case your data catalog is large, loading all partitions for all tables can be quite time consuming.
+    Skipping partition information can reduce your metastore refresh latency
+    """,
+)
--- a/querybook/server/lib/metastore/loaders/form_fileds.py
+++ b/querybook/server/lib/metastore/loaders/form_fileds.py
@@ -1,7 +1,7 @@
 from lib.form import FormField, FormFieldType
 
 
-load_partitions = FormField(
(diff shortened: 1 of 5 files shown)


querybook/server/lib/metastore/loaders/form_fileds.py
```diff
@@ -0,0 +1,10 @@
+from lib.form import FormField, FormFieldType
+
+
+load_partitions = FormField(
```
czgu, line 4: load_partitions_field

froukees: ok, renamed

czgu (approved):

Issue #691, opened 2021-11-05
LDAP auth - bind user, filter


froukees:
## Story
As a deployer of Querybook in an organization, it would be nice to have an option to configure a LDAP filter which would allow access to only specific group of LDAP users.

### Current state
Current [ldap_auth.py](https://github.com/pinterest/querybook/blob/master/querybook/server/app/auth/ldap_auth.py) implementation authenticates a user against LDAP and if it's successful, the user is straight away allowed to the WebUI.

### Proposal
Proposed solution would check if an user has a claim for an access based on a filter, then checked the credentials as follows:
1. use a _bind_ (technical) user for initialization of LDAP connection
2. use the _bind_ user for obtaining information about the current session user based on username and LDAP filter
    - if user doesn't comfort the filter (0 users returned) - raise `Unauthorized`
3. use the session username and password for LDAP authentication
    - if fails, raise `Bad credentials`
    - otherwise pass the user

#### Impact
The _bind_ user can be mandatory/breaking configuration, or it can be done in a way, that proposed logic is applied only in case a _bind_ user is configure.

PR #717, opened 2021-12-22
Feat: LDAP bind user and filter

froukees:
Solving #691 

This pull request:
- Extends `ldap_auth.py` module by possibility to specify LDAP bind user that can do LDAP search based on a filtering condition and allow/block specific users to access WebUI.
- Handles situations when user enters LDAP `dn` in different format than `uid=...`.
- Backward compatibility is ensured. Logging in directly with full LDAP DN or with uid and `LDAP_USER_DN` set is still possible.

I was stuck little bit on unit tests, because I based them on `mockldap`  library, but then found out it's not installable under Python 3.7.9 in the GitHub workflow, so I created custom mocked responses instead.

### DIFF
--- a/querybook/server/app/auth/ldap_auth.py
+++ b/querybook/server/app/auth/ldap_auth.py
@@ -1,72 +1,318 @@
-import re
+from contextlib import contextmanager
+from dataclasses import dataclass
+from typing import Dict, Iterator, List, Optional, Tuple, Union
 
 import ldap
 import flask_login
+import re
+from ldap.ldapobject import SimpleLDAPObject
 
 from flask_login import current_user
 from env import QuerybookSettings
 from app.db import with_session, DBSession
 from .utils import AuthenticationError, AuthUser, QuerybookLoginManager
+from lib.logger import get_logger
 from logic.user import (
     get_user_by_name,
     create_user,
 )
+from models.user import User
+
+LOG = get_logger(__file__)
 
 login_manager = QuerybookLoginManager()
 
 
-def get_transformed_username(username):
-    dn = None
-    if username.startswith("uid="):
-        dn = username
-
-        match = re.match(r"^uid=([^,]+)", username)
-        username = match.group(1)
-    else:
-        dn = QuerybookSettings.LDAP_USER_DN.format(username)
-    return username, dn
+class LDAPAuthErrors:
+    BAD_CREDENTIALS = "Bad credentials"
+    UNAUTHORIZED = "Unauthorized"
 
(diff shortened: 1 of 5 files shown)


querybook/server/app/auth/ldap_auth.py
```diff
@@ -1,76 +1,317 @@
...
+    # Remove any search referrals from results
+    cleared_results = [
+        (dn, attrs)
+        for dn, attrs in search_results
+        if dn is not None and isinstance(attrs, dict)
+    ]
+    n_found = len(cleared_results)
+    if n_found != 1:
+        if n_found > 1:
+            logger.warning("LDAP search returned %s results" % n_found)
```
czgu, line 58: use format string here?

querybook/server/app/auth/ldap_auth.py
```diff
@@ -1,76 +1,317 @@
...
+    first_name: Optional[str] = None
+    last_name: Optional[str] = None
+
+    @property
+    def full_name(self) -> str:
+        full_name = " ".join(filter(bool, [self.first_name, self.last_name]))
+        return full_name if full_name else self.cn
+
+
+def _clear_ldap_search_results(
```
czgu, line 46: _sanitize_ldap_search_results?

querybook/server/app/auth/ldap_auth.py
```diff
@@ -1,76 +1,317 @@
...
 
 login_manager = QuerybookLoginManager()
+logger = get_logger(__name__)
 
 
-def get_transformed_username(username):
-    dn = None
-    if username.startswith("uid="):
-        dn = username
+class AuthErrors:
```
czgu, line 25: LDAPAuthErrors

czgu (approved):
Great work! the code/documentation/testing is high quality 👍

querybook/server/app/auth/ldap_auth.py
froukees: Ok 👍

querybook/server/app/auth/ldap_auth.py
froukees: Ok 👍  I've aligned log messages to f-strings in this module. Also, I've changed `logger = get_logger(__name__)` to `LOG = get_logger(__file__)`, as it seems much more common in the repo.

querybook/server/app/auth/ldap_auth.py
froukees: Ok 👍

PR #728, opened 2022-01-12
fix: upgrade elasticsearch to 7.16.2 to fix log4j issue

kgopal492:
Context: https://www.elastic.co/blog/new-elasticsearch-and-logstash-releases-upgrade-apache-log4j2

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -31,7 +31,7 @@ pyyaml==5.4.1
 sqlalchemy==1.3.17
 pymysql==0.9.3
 requests==2.22.0
-elasticsearch==7.16.2
+elasticsearch==7.13.4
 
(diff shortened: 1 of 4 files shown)


requirements/base.txt
```diff
@@ -31,7 +31,7 @@ pyyaml==5.4.1
 sqlalchemy==1.3.17
 pymysql==0.9.3
 requests==2.22.0
-elasticsearch==7.13.4
+elasticsearch==7.16.2
```
czgu, line 34: can you please keep the version to be 7.13.4 since aws es7 no longer works after that version? context: https://github.com/pinterest/querybook/pull/685#issuecomment-975894514

czgu (approved):

PR #742, opened 2022-02-14
ui: add more accent

meowcodes:
Screen Shot 2022-02-14 at 2 15 02 PM


querybook/webapp/ui/Icon/Icon.tsx
```diff
@@ -10,10 +10,20 @@ export interface IIconProps {
     name: string;
     options?: FeatherAttributes;
     fill?: boolean;
+    color?: buttonColors;
 }
 
+export type buttonColors = 'accent' | 'true' | 'false' | 'warning';
```
czgu, line 16: use something like TButtonColors to make the typing standout

PR #747, opened 2022-02-19
Feat: Enable embedding of datadocs

darapuk:
Purpose:
This PR exposes an endpoint to enable embedding of DataDocs in external Applications. One such use case will be the ability to search and discover DataDocs through the open source project [Datahub](https://github.com/linkedin/datahub).

It is not certain that this is the method that we will employ to fully embed the DataDocs, but wanted to open up this workflow for the Querybook community if it would be useful.

External Dependencies:
Metadata modeling to ingest DataDocs to Datahub: https://datahubspace.slack.com/archives/C02R2NBJXD1/p1645047031441309

Requirements:
- [ ] Update Docs
- [x] Add Tests

### DIFF
--- a/querybook/webapp/components/EmbeddedDataDocPage/EmbeddedDataDocPage.tsx
+++ b/querybook/webapp/components/EmbeddedDataDocPage/EmbeddedDataDocPage.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import { FullHeight } from 'ui/FullHeight/FullHeight';
+import './EmbeddedDataDocPage.scss';
+import { DataDoc } from 'components/DataDoc/DataDoc';
+
+const EmbeddedDataDocPage: React.FunctionComponent = ({id}) => {
+    return (
+        <FullHeight flex={'column'} className="EmbeddedDataDocPage">
+            <div className="datadoc-wrapper">
+                <DataDoc docId={id} />
+            </div>
+        </FullHeight>
+    );
+};
+
+export default EmbeddedDataDocPage;
--- a/querybook/webapp/components/EmbeddedDataDocPage/EmbeddedDataDocPage.tsx
+++ b/querybook/webapp/components/EmbeddedDataDocPage/EmbeddedDataDocPage.tsx
@@ -2,7 +2,11 @@ import React from 'react';
 import { FullHeight } from 'ui/FullHeight/FullHeight';
 import { DataDoc } from 'components/DataDoc/DataDoc';
 
-const EmbeddedDataDocPage: React.FunctionComponent = ({ id }) => {
+interface IProps {
+    id: number;
+}
+
+const EmbeddedDataDocPage: React.FunctionComponent<IProps> = ({ id }) => {
     return (
         <FullHeight flex={'column'} className="EmbeddedDataDocPage">
             <div>
--- a/querybook/webapp/components/EmbeddedDataDocPage/EmbeddedDataDocPage.tsx
+++ b/querybook/webapp/components/EmbeddedDataDocPage/EmbeddedDataDocPage.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { FullHeight } from 'ui/FullHeight/FullHeight';
 import { DataDoc } from 'components/DataDoc/DataDoc';
 
-const EmbeddedDataDocPage: React.FunctionComponent = ({id}) => {
+const EmbeddedDataDocPage: React.FunctionComponent = ({ id }) => {
     return (
         <FullHeight flex={'column'} className="EmbeddedDataDocPage">
             <div>
--- a/querybook/webapp/components/EmbeddedDataDocPage/EmbeddedDataDocPage.tsx
+++ b/querybook/webapp/components/EmbeddedDataDocPage/EmbeddedDataDocPage.tsx
@@ -1,11 +1,14 @@
 import React from 'react';
 import { FullHeight } from 'ui/FullHeight/FullHeight';
+import './EmbeddedDataDocPage.scss';
 import { DataDoc } from 'components/DataDoc/DataDoc';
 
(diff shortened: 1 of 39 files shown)


querybook/webapp/components/EnvironmentAppRouter/EnvironmentModalSwitchRouter.tsx
```diff
@@ -125,6 +128,10 @@ export const EnvironmentModalSwitchRouter: React.FC = () => {
                         path="/:env/_/embedded_editor/"
                         render={() => <EmbeddedQueryPage />}
                     />
+                    <Route
+                        path="/:env/embedded_datadoc/:docId"
```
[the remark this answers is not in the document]
czgu: can you update publishDataDocTitle in DataDoc.tsx to account for `/_/embedded_datadoc/`.  Once we are done with `pathParts = location.pathname.split('/')`, you can check if pathParts[1] is `'_'` or `'datadoc'`

querybook/webapp/components/EnvironmentAppSidebar/EnvironmentAppSidebar.tsx
```diff
@@ -19,9 +20,12 @@ import { matchKeyMap, KeyMap } from 'lib/utils/keyboard';
 import './EnvironmentAppSidebar.scss';
 
 const SIDEBAR_WIDTH = 320;
+const EMBED_PATH_STRING = 'embedded';
 
 export const EnvironmentAppSidebar: React.FunctionComponent = () => {
-    const [collapsed, setCollapsed] = React.useState(false);
+    const location = useLocation();
+    const isEmbed = location.pathname.includes(EMBED_PATH_STRING);
+    const [collapsed, setCollapsed] = React.useState(isEmbed);
```
czgu, line 28: also i think its cleaner if you check `/_/` is in pathname instead.

```
React.useState(() => location.pathname.includes(EMBED_PATH_STRING));

File: querybook/webapp/components/EmbeddedDataDocPage/EmbeddedDataDocPage.tsx [hunk 3]
```diff
@@ -0,0 +1,19 @@
...
+import { DataDoc } from 'components/DataDoc/DataDoc';
+
+interface IProps {
+    id: number;
+}
+
+const EmbeddedDataDocPage: React.FunctionComponent<IProps> = ({ id }) => {
+    return (
+        <FullHeight flex={'column'} className="EmbeddedDataDocPage">
+            <div>
```
czgu, line 12: this <div> is not necessary?

czgu (changes requested):

querybook/webapp/components/EmbeddedDataDocPage/EmbeddedDataDocPage.tsx
darapuk: This gives the embedded datadoc the ability to scroll. Setting overflow on the DataDoc component didn't give the desired effect

querybook/webapp/components/EnvironmentAppSidebar/EnvironmentAppSidebar.tsx
darapuk: Updated

darapuk:
Screen Shot 2022-02-22 at 3 07 47 PM

czgu (approved):

querybook/webapp/components/EmbeddedDataDocPage/EmbeddedDataDocPage.tsx
darapuk: > Just to confirm, div is needed for DataDoc scroll, right?
Yup, exactly

darapuk:
Could I get one more review of this @czgu?

querybook/webapp/redux/querybookUI/reducer.ts
```diff
@@ -10,7 +11,7 @@ const initialState: IQuerybookUIState = {
         recent: true,
         favorite: true,
     },
-    isEnvCollapsed: false,
+    isEnvCollapsed: false || isEmbedPath(window.location.pathname),
```
czgu, line 14: same as `isEnvCollapsed: isEmbedPath(window.location.pathname),`

querybook/webapp/lib/utils/embed.ts
```diff
@@ -0,0 +1,4 @@
+const EMBED_PATH_STRING = 'embedded';
```
czgu, line 1: this is a bit too relaxed? since you could have datadoc that has embedded in the title and that would modify the url to have embedded as string. It is better to check if the path contains `/_/embedded`

querybook/webapp/lib/utils/embed.ts
```diff
@@ -0,0 +1,4 @@
...
+
+export const isEmbedPath = (path: string): boolean =>
```
czgu, line 3: imo this is a really one off case that we don't need a file for it, since it could be done with `window.location.pathname.includes(EMBED_PATH_STRING)`

querybook/webapp/redux/querybookUI/reducer.ts
darapuk: 🤦🏽‍♂️

querybook/webapp/lib/utils/embed.ts
darapuk: Good point, updated

querybook/webapp/lib/utils/embed.ts
darapuk: Ya, I was trying to avoid adding logic into the reducer, but you're right, it's pretty bespoke so it's probably fine to leave here.

Can always refactor later if we add different embedding capabilities.

czgu (approved):

PR #753, opened 2022-02-26
feat: Add templating guide, make templated variable embedded in doc

czgu:
Editable empty
image

Editable with variables
image

Readonly With variables
image

Readonly Empty
image

Tutorial guide
image

Modal form
image

### DIFF
--- a/querybook/webapp/ui/Modal/Modal.scss
+++ b/querybook/webapp/ui/Modal/Modal.scss
@@ -5,8 +5,8 @@
     }
 
     .Modal-top {
-        font-family: var(--font-title);
         .Modal-title {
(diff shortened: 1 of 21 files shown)


querybook/webapp/ui/Modal/Modal.scss
```diff
@@ -5,8 +5,8 @@
     }
 
     .Modal-top {
-        font-family: var(--font-title);
         .Modal-title {
+            font-family: var(--font-title);
```
meowcodes, line 9: can you keep it on top so it applies to other things in modal top like tabs? lets just force all buttons to be font-sans-serif

meowcodes (approved):

PR #755, opened 2022-03-01
feat: More UI Updates

czgu:
- Make boards filter case insensitive
- Sample query filters are now hidden by default, latest partition is auto selected
image
image
- Editable Text Field auto focus
- Modal close after snippet insert

Fixes:
- TaskEditor can now use recurrence editor for more cron expressions
- Tooltip should have a higher z-index to prevent it being overlapped

### DIFF
--- a/querybook/webapp/components/Task/TaskEditor.tsx
+++ b/querybook/webapp/components/Task/TaskEditor.tsx
@@ -310,7 +310,7 @@ export const TaskEditor: React.FunctionComponent<IProps> = ({
         );
 
         const canUseRecurrence = validateCronForRecurrrence(values.cron);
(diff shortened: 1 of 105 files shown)


querybook/webapp/components/Task/TaskEditor.tsx
```diff
@@ -310,7 +310,7 @@ export const TaskEditor: React.FunctionComponent<IProps> = ({
         );
 
         const canUseRecurrence = validateCronForRecurrrence(values.cron);
-
+        console.log({ values, canUseRecurrence });
```
kgopal492, line 313: remove?

kgopal492 (approved):

Issue #756, opened 2022-03-02
Generic OAuth fails after redirection


thalesbrant:
Steps to reproduce

1 - Setup a generic OAuth provider (i.e: Keycloak) with a user (i.e: admin/admin)
2 - Follow the steps to configure a generic OAuth from Querybook (https://www.querybook.org/docs/configurations/infra_config#authentication)
3 - Disable HOT-RELOAD by setting the environment variable to false in the **docker-compose.yml** file.
4 - Start the systems and try to open the Querybook UI in incognito mode
5 - If everything is configured properly, you shall see Keycloak UI for the login step
6 - After login, the redirection to Querybook UI fails by showing up an error message regarding Unauthorized Access as shown below

- _The server could not verify that you are authorized to access the URL requested. You either supplied the wrong credentials (e.g. a bad password), or your browser doesn't understand how to supply the credentials required._

> After analyzing the code, it was found that the request to fetch the user details is done via params argument only with no possibility to change that to be used in headers.

Expected
OAuth login is succeded and Querybook UI is shown properly.

PR #757, opened 2022-03-02
feat: Generic OAuth - Give the possibility to use the access_token in headers section

thalesbrant:
Solving #756

This PR gives flexibility to the Generic OAuth Flow when using a provider that accepts authorized API calls using the retrieved access_token via Authorization header instead of URL parameters.

### DIFF
--- a/docs_website/docs/configurations/infra_config.md
+++ b/docs_website/docs/configurations/infra_config.md
@@ -98,8 +98,6 @@ the next few configurations are only relevant if you are using OAuth based authe
 -   `OAUTH_AUTHORIZATION_URL` (**required**): Url for oauth redirection
 -   `OAUTH_TOKEN_URL` (**required**): Url to get the oauth token
 -   `OAUTH_USER_PROFILE` (**required**): Url to get the user profile
--   `OAUTH_TOKEN_PLACE`: Place to send the obtained access_token (headers) or default if not informed (url params)
--   `OAUTH_TOKEN_HEADER_PREFIX`: Prefix to add to the Authorization header (i.e: Bearer)
 
 for LDAP authentication:
 
--- a/docs_website/docs/configurations/infra_config.md
+++ b/docs_website/docs/configurations/infra_config.md
@@ -98,6 +98,8 @@ the next few configurations are only relevant if you are using OAuth based authe
 -   `OAUTH_AUTHORIZATION_URL` (**required**): Url for oauth redirection
 -   `OAUTH_TOKEN_URL` (**required**): Url to get the oauth token
 -   `OAUTH_USER_PROFILE` (**required**): Url to get the user profile
+-   `OAUTH_TOKEN_PLACE`: Place to send the obtained access_token (headers) or default if not informed (url params)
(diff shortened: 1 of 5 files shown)


docs_website/docs/configurations/infra_config.md
```diff
@@ -98,6 +98,8 @@ the next few configurations are only relevant if you are using OAuth based authe
 -   `OAUTH_AUTHORIZATION_URL` (**required**): Url for oauth redirection
 -   `OAUTH_TOKEN_URL` (**required**): Url to get the oauth token
 -   `OAUTH_USER_PROFILE` (**required**): Url to get the user profile
+-   `OAUTH_TOKEN_PLACE`: Place to send the obtained access_token (headers) or default if not informed (url params)
```
czgu, line 101: OAUTH_PROFILE_TOKEN_PLACE

docs_website/docs/configurations/infra_config.md
```diff
@@ -98,6 +98,8 @@ the next few configurations are only relevant if you are using OAuth based authe
...
+-   `OAUTH_TOKEN_HEADER_PREFIX`: Prefix to add to the Authorization header (i.e: Bearer)
```
czgu, line 102: Hm do you know if there is any case where the format is not "authorization: bearer <token>"?This option might not be necessary?

querybook/server/app/auth/oauth_auth.py
```diff
@@ -109,9 +111,19 @@ def _fetch_access_token(self, code):
         return resp["access_token"]
 
     def _get_user_profile(self, access_token):
-        resp = requests.get(
-            self.oauth_config["profile_url"], params={"access_token": access_token}
-        )
+        if(self.oauth_config["token_place"] == 'headers'):
```
czgu, line 114: can we make headers the default option actually? I doubt any oauth provider uses access_token from query params. This was legacy code from pinterest's internal oauth system that we didn't bother to fix 😢

czgu (commented):
Thanks for fixing this!

thalesbrant:
So, after your comments, I think we can make it by default to be in the header. 
I don't have any other case where we could use a different authorization prefix beyond 'Bearer'. I just made it flexible because I did not know if it would break anyone's integration.

docs_website/docs/configurations/infra_config.md
thalesbrant: Removed in the last commit. By default, it would be in the headers.

docs_website/docs/configurations/infra_config.md
thalesbrant: Removed in the last commit. By default, it would be in the headers.

querybook/server/app/auth/oauth_auth.py
thalesbrant: Removed in the last commit. By default, it would be in the headers.

thalesbrant:
Version is bumped, lint fixed and breaking_changes note added.

czgu (approved):

czgu:
Great thank for working on this!

PR #779, opened 2022-03-21
Add hot keys for commenting and re-formatting lines

alexVasylenko:
This code should improve the hot keys for query formatter.

when i am on querybook in the online editor, and i am working on code within a cell block, and i have some code highlighted
the commenting out standard I have experience in to comment and uncomment code is control forward slash (ctr + /)
word wrapped lines of text are treated as a single line.

when i am on querybook in the online editor, and i am working on code within a cell block
the code shifting standard I have experience in to move code up and down is alt up arrow key and alt down arrow key (alt + ↑, alt + ↓)
word wrapped lines of text are treated as a single line


czgu:
hey Alex, shortcut should be configured in keyMap.ts for standardization. For the two you are adding, I have some questions:
- block-comment is already provided, does it not work on Windows? 
- It is a good idea to add swap lines, since it is provided in codemirror sublime keymap, I just copied the code there and added it in this PR https://github.com/pinterest/querybook/pull/782.

alexVasylenko:
Hi! I have updated the MR. So, the main idea for this commit was to add a new shortcut (Ctrl+ /) for comment/uncomment string. The current editor works fine with (Cmd+/) but Ctrl is more natural for Windows users.

czgu:
@alexVasylenko ah got it, it is actually because codemirror doesn't support `cmd+/` for windows. Originally I imagined it would auto use ctrl in that case. I added a PR here for auto coversion and https://github.com/pinterest/querybook/pull/793 and now `ctrl+/` should work out of the box.

PR #781, opened 2022-03-21
ui: remove fontawesome/feather, use lucide

czgu:
- No visible UI change
- Lucide is used to replace Feather, also added better IconValidation and tree shaking
- Fixed some UI warnings when loading Querybook


meowcodes (approved):
this pr makes me happy

PR #805, opened 2022-03-24
Add drag n drop for database names

alexVasylenko:
As a QueryBook user I would like a option when you can insert table names to the query interfaces (adhoc/Datadoc) to improve efficiency from browsing tables to querying them.
This commit is added a drag and drop feature to querybook.
Снимок экрана от 2022-03-24 17-51-10


czgu (changes requested):
Link is a generic URL component, don't tie in business logic into it

PR #810, opened 2022-03-25
fix: CSV iterator logic update

czgu:
Previously, we assumed each line is a CSV row. But this might not be the case since CSV cells could also contain new line characters. In the new logic, the CSV reader would parse the row into valid CSV chunks and then return


querybook/server/lib/result_store/stores/db_store.py
```diff
@@ -32,9 +30,10 @@ def _get_first_n_lines(self, n: Optional[int]) -> List[str]:
     def get_csv_iter(
         self, number_of_lines: Optional[int] = None
     ) -> Generator[List[List[str]], None, None]:
-        curr_lines = self._get_first_n_lines(number_of_lines)
-        raw_csv_str = "\n".join(curr_lines)
-        return str_to_csv_iter(raw_csv_str)
+        csv_iter = str_to_csv_iter(self.read_raw())
+        if number_of_lines is None:
```
kgopal492, line 34: I think if the second arg of `islice` is None, it'll return the entire iterator, so this if statement can be removed

querybook/server/lib/utils/csv.py
```diff
@@ -0,0 +1,140 @@
...
+                )
+
+            columns_count = 0
+            last_valid_index = line_idx
+
+    return last_valid_index
+
+
+def split_csv_to_chunks(lines: List[str]) -> Tuple[List[str], List[str]]:
+    """Given a serials of lines, break them into 2 parts
```
kgopal492, line 122: nit: "series of lines" or "list of lines"

kgopal492 (approved):

kgopal492 (approved):

PR #834, opened 2022-04-12
Opportunity to redefine uid attribute (fixes #833)

### LINKED ISSUE #833: Opportunity to redefine uid attribute in LDAP authorization
**Problem:**
_uid_ attribute is hardcoded in filters and attributes in _ldap_auth.py_ right now, but some comanies use different identifying attributes in their Active Directories.

**Solution:**
Add _LDAP_UID_FIELD_ setting and use it everywhere.

a-pertsev:


querybook/server/env.py
```diff
@@ -14,9 +14,9 @@ class MissingConfigException(Exception):
     pass
 
 
-def get_env_config(name, optional=True):
+def get_env_config(name, optional=True, default=None):
```
czgu, line 17: for default values, can you put it in querybook_default_config.yaml? there is no need to put defaults in env.py

czgu (commented):
just a minor comment, can you also put fixes #833 in the description of the PR so it can also link to the issue?

querybook/server/env.py
a-pertsev: done!

PR #836, opened 2022-04-12
ui: dag exporter modal

meowcodes:
Screen Shot 2022-04-11 at 7 45 51 PM


querybook/webapp/lib/sql-helper/sql-formatter.ts
```diff
@@ -169,3 +169,15 @@ function getLanguageForSqlFormatter(language: string): SqlFormatterLanguage {
 
     return 'sql';
 }
+
+export const getQueryStatements = (query: string) => {
+    const statements = new Set();
```
czgu, line 174: given this is a place holder, can you do:

```
export const getQueryStatements = (query: string) => {
    return ['SELECT', 'UPDATE'];
}

File: querybook/webapp/ui/FlowGraph/LineageNode.tsx [hunk 2]
```diff
@@ -49,6 +49,11 @@ export const LineageNode = React.memo<IProps>(({ data }) => {
                 position={Position.Right}
                 isConnectable={false}
             />
+            <Handle
```
czgu, line 52: duplicate

PR #843, opened 2022-04-18
Pass utf8 as default charset into mysqld command (fixes #842)

### LINKED ISSUE #842: Use unicode as default charset for mysql database
Problem:

If user fullname contains non latin characters (for example, cyrillic), there is an excaption during user creation
```
(pymysql.err.InternalError) (1366, "Incorrect string value: '\\xD0\\x90\\xD0\\xBB\\xD0\\xB5...' for column 'fullname' at row 1") 
```
Proposal:
Pass utf8 as mysql default charset.

a-pertsev:
Pass utf8 as default charset into mysqld command (fixes #842)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -95,6 +95,7 @@ services:
             interval: 30s
             timeout: 30s
             retries: 3
+        command: ['--character-set-server=utf8mb4', '--collation-server=utf8mb4_unicode_ci']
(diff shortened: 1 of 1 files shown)


a-pertsev:
@czgu could you look at this PR, please?

czgu (approved):
thanks!

PR #868, opened 2022-05-17
feat: Add ability to customize tags

czgu:
Add ability to annotate tags with the following features:
- Color
- Tooltip
- Icon
- Rank
- Admin (Only admin can edit)

image

You can edit tags by right clicking (only the ones you are allowed to)
image
image


querybook/webapp/ui/Tag/Tag.tsx
```diff
@@ -27,16 +26,42 @@ export interface ITagProps {
...
 
 export const TagGroup: React.FunctionComponent<ITagGroupProps> = ({
     className,
     children,
 }) => <div className={`${className} TagGroup`}>{children}</div>;
 
+const StyledColorTag = styled.span.attrs<{
+    highlighted?: boolean;
+    light?: boolean;
+    color?: string;
```
meowcodes, line 38: are we allowing all colors? i dont think we should - lets limit to the colors we have red, blue, yellow, pink, green, etc. and use the -light or -lightest for bg and -dark or -darkest for text

meowcodes (approved):

PR #874, opened 2022-05-25
feat: Allow metastore loaders to provide table/column descriptions

### LINKED ISSUE #871: Importing Table Metadata from an External System
Hi, we are looking at how we can maintain table metadata (e.g. table descriptions, column descriptions, table warnings) in a separate system, but mirror it into Querybook.  It doesn't look like it's possible currently, so I've come up with a few options:

1. Write an automated script that exports metadata and pushes it into Querybook database.  This doesn't require changes to Querybook, but relying on Querybook internals is brittle.

2. Extend one of the Metastore plugins (e.g. `HMSMetastoreLoader`) to fetch data during load.   This would only require a new plugin, so no changes in Querybook.  Downside is that it would be implemented per metastore, so if we wanted to implement this for a variety of different metastores, it would be a bit messy.  It would also require maintaining a separate plugin fork.

3. Extend Querybook with a new plugin type that pulls table metadata from an external source.  This could be triggered as part of the `update_metastore` task, or configured as a separate task that runs on its own schedule.  This allows the implementation to be cleanly separated from both the Querybook internals and any specific metastore loader.

I think the 3rd option is the best approach, but requires the most changes in Querybook to support.  Does this sound like a reasonable approach? Are there any other solutions that I'm missing?

baumandm:
This change adds description properties to `DataTable` and `DataColumn`, and persists these to the database (if not `None`).

This allows a metastore loader implementation to source the table and column descriptions from a separate system and sync them into Querybook.

Here's an example implementation that hardcodes the same descriptions for every table & column:

```python
from typing import List, Tuple
from lib.logger import get_logger
from lib.metastore.base_metastore_loader import DataColumn, DataTable
from lib.metastore.loaders.sqlalchemy_metastore_loader import SqlAlchemyMetastoreLoader

LOG = get_logger(__name__)

class ExternalMetadataLoader(SqlAlchemyMetastoreLoader):
    def get_table_and_columns(
        self, schema_name, table_name
    ) -> Tuple[DataTable, List[DataColumn]]:
        table, columns = super().get_table_and_columns(schema_name, table_name)

        table = table._replace(description="test table description")

        columns = list(
            map(
                lambda column: column._replace(description="test column description"),
                columns,
            )
        )
        
        return table, columns
```

Fixes #871.

### DIFF
--- a/querybook/server/logic/metastore.py
+++ b/querybook/server/logic/metastore.py
@@ -255,6 +255,7 @@ def create_table_information(
     data_table_id=None,
     latest_partitions=None,
     earliest_partitions=None,
+    description=None,
     hive_metastore_description=None,
     commit=False,
     session=None,
@@ -269,6 +270,9 @@ def create_table_information(
         earliest_partitions=earliest_partitions,
         hive_metastore_description=hive_metastore_description,
     )
+
+    if description:
+        new_table_information.description = description
 
     if not table_information:
         session.add(new_table_information)
@@ -419,7 +423,13 @@ def get_all_column_name_by_table_id(table_id, session=None):
 
 @with_session
 def create_column(
-    name=None, type=None, comment=None, table_id=None, commit=True, session=None
+    name=None,
+    type=None,
+    comment=None,
+    description=None,
+    table_id=None,
+    commit=True,
+    session=None,
 ):
     old_table_column = get_column_by_name(name, table_id, session=session)
     if old_table_column:
@@ -431,6 +441,9 @@ def create_column(
         comment=comment,
         table_id=table_id,
     )
+
+    if description:
+        new_table_column.description = description
 
     if not old_table_column:
         session.add(new_table_column)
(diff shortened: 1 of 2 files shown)


querybook/server/logic/metastore.py
```diff
@@ -270,6 +271,9 @@ def create_table_information(
         hive_metastore_description=hive_metastore_description,
     )
 
+    if description:
+        new_table_information.description = description
```
jczhong84, line 275: what if the description from metastore is an empty string ""? may want to check it by
`if description is not None`

querybook/server/logic/metastore.py
```diff
@@ -432,6 +442,9 @@ def create_column(
         table_id=table_id,
     )
 
+    if description:
```
jczhong84, line 445: same here

querybook/server/logic/metastore.py
LulaV14: In Python empty strings are considered falsy, so this if statement should cover that case

jczhong84: what I meant was if the description is an empty string from metastore, we would like to honor it and load the empty description to querybook to overwrite the old value.

baumandm: Good catch, yes that is a good use case.

baumandm:
Closing this in favor of #980

PR #875, opened 2022-05-26
fix: restrict access to private queries in query search

### LINKED ISSUE #865: Private DataDocs are not really Private
DataDoc security and Query Execution visibility highlight a hole in the feature of a Private DataDoc.

**The use case is as follows:**

- User creates DataDoc and flags this as Private 
- Another non authorized user of the DataDoc can 'search' with a keyword in the DataDoc search feature.  The result returns the DataDoc title & the actual query within. 
DataDocSearch

- When clicking on the DataDoc they get access denied which is right.
AccessDenied

- More scary is if they use the right search term ie something from the Query they can find the execution from the DataDoc and open that Execution as well as see the results of that.
QuerySearch
Execution

kgopal492:
fixes #865

- add `public` and `readable_user_ids` fields to query_cell and query_execution elasticsearch indices to restrict access to queries on private datadocs
- change elasticsearch item creation functions to allow option to generate/update specific fields
- update `querybook/server/datasources/search.py` to use "must" query to ensure user has access to datadocs & queries that are returned

### DIFF
--- a/querybook/server/datasources/search.py
+++ b/querybook/server/datasources/search.py
@@ -105,6 +105,14 @@ def _match_filters(filters):
     return filters
 
 
+def _query_access_terms(user_id):
+    return [
+        {"term": {"author_uid": user_id}},
+        {"term": {"readable_user_ids": user_id}},
+        {"term": {"public": True}},
+    ]
+
+
 def _construct_query_search_query(
     keywords,
     filters,
@@ -148,6 +156,12 @@ def _construct_query_search_query(
         search_query = {"match_all": {}}
 
     search_filter = _match_filters(filters)
+    if search_filter == {}:
+        search_filter["filter"] = {"bool": {}}
+    search_filter["filter"]["bool"].setdefault("must", [])
+    search_filter["filter"]["bool"]["must"].append(
+        {"bool": {"should": _query_access_terms(current_user.id)}}
+    )
 
     bool_query = {}
     if search_query != {}:
@@ -213,7 +227,10 @@ def _construct_datadoc_query(
     search_filter = _match_filters(filters)
     if search_filter == {}:
         search_filter["filter"] = {"bool": {}}
-    search_filter["filter"]["bool"]["should"] = _data_doc_access_terms(current_user.id)
+    search_filter["filter"]["bool"].setdefault("must", [])
+    search_filter["filter"]["bool"]["must"].append(
+        {"bool": {"should": _data_doc_access_terms(current_user.id)}}
+    )
 
     bool_query = {}
     if search_query != {}:
--- a/querybook/server/datasources/search.py
+++ b/querybook/server/datasources/search.py
@@ -158,8 +158,7 @@ def _construct_query_search_query(
     search_filter = _match_filters(filters)
     if search_filter == {}:
         search_filter["filter"] = {"bool": {}}
-    search_filter["filter"]["bool"].setdefault("must", [])
-    search_filter["filter"]["bool"]["must"].append(
+    search_filter["filter"]["bool"].setdefault("must", []).append(
         {"bool": {"should": _query_access_terms(current_user.id)}}
     )
 
@@ -227,8 +226,7 @@ def _construct_datadoc_query(
     search_filter = _match_filters(filters)
     if search_filter == {}:
         search_filter["filter"] = {"bool": {}}
-    search_filter["filter"]["bool"].setdefault("must", [])
-    search_filter["filter"]["bool"]["must"].append(
+    search_filter["filter"]["bool"].setdefault("must", []).append(
         {"bool": {"should": _data_doc_access_terms(current_user.id)}}
     )
 
--- a/querybook/server/logic/elasticsearch.py
+++ b/querybook/server/logic/elasticsearch.py
@@ -92,10 +92,13 @@ def _get_dict_by_field(
         Dict[str, Any] - subset of the fields passed or the entire generated dict if fields is None
     """
     fields = fields or field_to_getter.keys()
+
+    def get_field_value(field: str):
+        getter = field_to_getter[field]
+        return getter() if callable(getter) else getter
+
     return {
-        field: getter() if callable(getter) else getter
-        for field, getter in field_to_getter.items()
-        if field in fields
+        field: get_field_value(field) for field in fields if field in field_to_getter
     }
 
 
--- a/querybook/server/logic/elasticsearch.py
+++ b/querybook/server/logic/elasticsearch.py
@@ -211,7 +211,7 @@ def query_execution_to_es(query_execution, data_cell=None, fields=None, session=
             else None
         )
 
-    field_to_generator = {
+    field_to_getter = {
         "id": query_execution.id,
         "query_type": "query_execution",
         "title": data_cell.meta.get("title", "Untitled") if data_cell else None,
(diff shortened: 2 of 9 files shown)


querybook/server/datasources/search.py
```diff
@@ -148,6 +156,12 @@ def _construct_query_search_query(
         search_query = {"match_all": {}}
 
     search_filter = _match_filters(filters)
+    if search_filter == {}:
+        search_filter["filter"] = {"bool": {}}
+    search_filter["filter"]["bool"].setdefault("must", [])
```
czgu, line 161:
```
search_filter["filter"]["bool"].setdefault("must", []).append(
        {"bool": {"should": _query_access_terms(current_user.id)}}
    )
```

querybook/server/datasources/search.py
```diff
@@ -213,7 +227,10 @@ def _construct_datadoc_query(
     search_filter = _match_filters(filters)
     if search_filter == {}:
         search_filter["filter"] = {"bool": {}}
-    search_filter["filter"]["bool"]["should"] = _data_doc_access_terms(current_user.id)
+    search_filter["filter"]["bool"].setdefault("must", [])
```
czgu, line 230: ditto

querybook/server/logic/elasticsearch.py
```diff
@@ -78,13 +80,19 @@ def get_hosted_es():
     return hosted_es
 
 
+def _get_partial_dict(field_to_generator, fields=None):
+    if fields is None:
```
czgu, line 84:
```py
fields = fields or field_to_generator.keys()
return {field: field_to_generator[field]() for field in fields if field in field_to_generator}

File: querybook/server/logic/elasticsearch.py [hunk 4]
```diff
@@ -78,13 +80,19 @@ def get_hosted_es():
     return hosted_es
 
 
+def _get_partial_dict(field_to_generator, fields=None):
```
czgu, line 83: def _get_dict_by_field(field_to_getter: Dict[str, Any] , fields: Optional[List[str]] =None):

also can you add documentation for this function

querybook/server/logic/elasticsearch.py
```diff
@@ -150,51 +159,66 @@ def _get_adhoc_query_executions_iter(batch_size=1000, session=None):
...
+        batch_size=batch_size, fields=fields, session=session
+    )
+    yield from _get_adhoc_query_executions_iter(
+        batch_size=batch_size, fields=fields, session=session
+    )
 
 
 @with_session
-def query_execution_to_es(query_execution, data_cell=None, session=None):
+def _get_datadoc_editors(datadoc, session=None):
```
czgu, line 172: remove the `@with_session` since session is "required" (datadoc must be bound to session)

querybook/server/logic/elasticsearch.py
```diff
@@ -150,51 +159,66 @@ def _get_adhoc_query_executions_iter(batch_size=1000, session=None):
...
+    """Return datadoc editors' uids. If there is no datadoc or datadoc is public, return empty array
```
czgu, line 173: use typing instead:
`def _get_datadoc_editors(datadoc, session=None) -> List[int]:`

querybook/server/logic/elasticsearch.py
```diff
@@ -150,51 +159,66 @@ def _get_adhoc_query_executions_iter(batch_size=1000, session=None):
...
-        "full_table_name": table_names,
-        "query_text": query_execution.query,
+    def get_table_names():
+        table_names, _ = lineage_lib.process_query(
+            query_execution.query, language=(engine and engine.language)
+        )
+        return list(chain.from_iterable(table_names))
+
+    field_to_generator = {
+        "id": lambda: query_execution.id,
```
czgu, line 206: upon second thought, there is no need to make every field a function since its slightly hard to read, there is a python method to check if a variable is a function https://stackoverflow.com/questions/624926/how-do-i-detect-whether-a-python-variable-is-a-function

querybook/server/logic/elasticsearch.py
```diff
@@ -263,33 +298,37 @@ def get_query_cells_iter(batch_size=1000, session=None):
...
-        "author_uid": datadoc and datadoc.owner_uid,
-        "engine_id": engine_id,
-        "statement_type": get_table_statement_type(query),
-        "created_at": DATETIME_TO_UTC(query_cell.created_at),
-        "full_table_name": table_names,
-        "query_text": query,
+    engine_id = query_cell_meta.get("engine")
+    engine = admin_logic.get_query_engine_by_id(engine_id, session=session)
+
+    def get_table_names():
```
czgu, line 309: can you make get_table_names into a helper function that is shared by execution and cell

querybook/server/logic/elasticsearch.py
```diff
@@ -728,3 +803,31 @@ def get_es_config_by_name(*config_names):
...
+        index_name = config["index_name"]
+        mapping = config["mappings"]["mappings"]
+        get_hosted_es().indices.put_mapping(mapping, index=index_name)
+
+
+def bulk_update_index_by_fields(config_name, fields):
+    es_config = ES_CONFIG[config_name]
+    type_name = es_config["type_name"]
+    if type_name == "query_executions":
+        LOG.info("Updating query executions")
```
czgu, line 820: LOG.info(f"Updating {type_name}")

querybook/tests/test_lib/test_elasticsearch/test_elasticsearch.py
```diff
@@ -0,0 +1,478 @@
...
+)
+
+CREATED_AT_DT = datetime.datetime(2020, 5, 5, 12, 8, 30)
+CREATED_AT_EPOCH = 1588680510
+COMPLETED_AT_DT = datetime.datetime(
+    2020, 5, 5, 12, 8, 45
+)  # 15 seconds after created at
+
+
+class BaseQueryTestCase(TestCase):
```
czgu, line 21: QueryTestCaseMixin

querybook/tests/test_lib/test_elasticsearch/test_elasticsearch.py
```diff
@@ -0,0 +1,478 @@
...
+            "title": self.QUERY_CELL_TITLE,
+            "data_doc_id": self.DATADOC_ID,
+            "environment_id": self.ENVIRONMENT_ID,
+            "author_uid": self.AUTHOR_UID,
+            "engine_id": self.ENGINE_ID,
+            "statement_type": ["SELECT"],
+            "created_at": CREATED_AT_EPOCH,
+            "full_table_name": ["main.test_table", "main.test_table_2"],
+            "query_text": self.QUERY_TEXT,
+            "public": False,
```
czgu, line 127: there is a lot of repetition, can we have some kind of shared dictionary for the common fields?

czgu (commented):
great work! minor style comments

querybook/server/logic/elasticsearch.py
kgopal492: Slightly changed to the following to check whether getter is a callable or value:

```
fields = fields or field_to_getter.keys()
return {
    field: getter() if callable(getter) else getter
    for field, getter in field_to_getter.items()
    if field in fields
}
```

czgu: super minor, but
```
for field, getter in field_to_getter.items()
    if field in fields
``` 
is n^2, so its faster to check `if field in field_to_getter`

kgopal492: ooh good point, fixed that!

czgu (dismissed):

czgu (dismissed):

czgu (approved):

PR #899, opened 2022-06-22
refactor: elasticsearch clean up

czgu:
- Breakdown elasticsearch logic into multiple files
- remove the assumption of running the code at the api level

### DIFF
--- a/querybook/server/lib/elasticsearch/search_datadoc.py
+++ b/querybook/server/lib/elasticsearch/search_datadoc.py
@@ -1,0 +1,89 @@
+from lib.elasticsearch.query_utils import (
+    match_any_field,
+    match_filters,
+    highlight_fields,
+)
+
+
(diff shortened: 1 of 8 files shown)


querybook/server/lib/elasticsearch/search_datadoc.py
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    if sort_key:
+        if not isinstance(sort_key, list):
+            sort_key = [sort_key]
+            sort_order = [sort_order]
+        sort_query = [
+            {val: {"order": order}} for order, val in zip(sort_order, sort_key)
+        ]
+
```
kgopal492, line 75: get_sort_query could probably be pulled out to a separate function, looks like the logic is same everywhere

kgopal492 (approved):

PR #929, opened 2022-07-08
ui: add list tour

meowcodes:


querybook/webapp/components/UIGuide/BoardItemUIGuide.tsx
```diff
@@ -0,0 +1,66 @@
...
+                button below the descriptions.
+            </p>
+        ),
+    },
+];
+
+export const BoardItemUIGuide: React.FunctionComponent = () => {
+    const location = useLocation();
+    const { tour } = React.useMemo(() => getQueryString(), [location.search]);
+    const [isOpen, setIsOpen] = React.useState(false);
```
czgu, line 45: its fine to make it a 1 time thing since they arent url triggered, like
```
const [isOpen, setIsOpen] = React.useState(() => !!getQueryString().tour);
```

czgu (commented):
what do you think of adding both boardUIguide and boarditemguide on the right sidebar instead? that way they will always be visible instead of showing up with 0,1 items?

czgu (approved):

PR #937, opened 2022-07-20
feat: Collapse and expand ad hoc query execution results

baumandm:
Previously, query executions could be closed on the ad hoc page, but could not be reopened without re-executing the query. This change allows collapsing and expanding the query execution results, so it behaves similar to the sidebar.


czgu (approved):

czgu:
great, thanks for adding this feature!

PR #939, opened 2022-07-21
feat: Add templating support for Adhoc query

jczhong84:
This PR is to add templating support for Adhoc query, like the datadocs. Changes include

- Add a dropdown menu to add template variables and preview template rendered query. 
- Also moved the "create datadoc" and "add udf" to the dropdown menu
- When creating datadoc from the adhoc query, it will pass the template variables over as well

image
image
image
image

PR #944, opened 2022-07-22
fix: adhoc template query

jczhong84:
Seeing below errors on adhoc queries. 
```
Failed to templatize query.
get_templated_query() missing 1 required positional argument: 'variables'
```
Cause:
In https://github.com/pinterest/querybook/pull/939, we added templating support for adhoc queries, while forgot to provide a default value when getting the `templatedVariables` from Redux store if there is no variables added, which will be undefined, while the render template API requires this argument.


querybook/webapp/components/QueryComposer/QueryComposer.tsx
```diff
@@ -12,7 +12,6 @@ import { useDispatch, useSelector } from 'react-redux';
 
 import { DataDocTemplateInfoButton } from 'components/DataDocTemplateButton/DataDocTemplateInfoButton';
 import { DataDocTemplateVarForm } from 'components/DataDocTemplateButton/DataDocTemplateVarForm';
-import { TTemplateVariableDict } from 'components/DataDocTemplateButton/helpers';
```
jczhong84, line 15: This is not used

czgu (approved):

PR #946, opened 2022-07-25
feat: add confirmation before execution if the query drops any tables

jczhong84:
**Context**
Recently one of our P0 prod tables were accidentally dropped by a user through querybook. At best we would like to prevent such cases from happening. At a minimum we will like an option to warn the user that they are about to drop a critical prod table and they need to confirm before the statement execution goes through.

This PR will ask user to confirm before the query execution if it will drop any tables.
Snip20220724_26

### DIFF
--- a/querybook/webapp/__tests__/lib/sql-helper/sql-checker.test.ts
+++ b/querybook/webapp/__tests__/lib/sql-helper/sql-checker.test.ts
@@ -1,0 +1,48 @@
+import { getDroppedTables } from 'lib/sql-helper/sql-checker';
+
+describe('getDroppedTables', () => {
+    test('drops a table', () => {
+        const query = `
+        drop table db.table1;
+        drop table db.table2;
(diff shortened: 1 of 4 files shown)


querybook/webapp/__tests__/lib/sql-helper/sql-checker.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        expect(getDroppedTables(query)).toStrictEqual([
+            'db.table1',
+            'db.table2',
+        ]);
+    });
+
+    test('drops a table and recreated', () => {
+        const query1 = `
+        drop table if exists db.table1;
+        create table db.table1 if not exists;
```
czgu, line 29: https://prestodb.io/docs/current/sql/create-table.html
create table [if not exists]

czgu (approved):

PR #948, opened 2022-07-27
fix: getStatementType vs invalid queries

czgu:
getDroppedTables fails silently if the query is invalid because getStatementType would throw an exception when there is no keywords in the query. This should no longer be the case as it would return null instead


jczhong84 (approved):
Thanks for fixing it!

PR #957, opened 2022-08-04
feat: dont show stack trace is exception is recognitizable

czgu:
There is no need for stack trace of QueryExecutionError since they are thrown by Querybook for runtime verification and can be traced

### DIFF
--- a/querybook/server/lib/query_executor/base_executor.py
+++ b/querybook/server/lib/query_executor/base_executor.py
@@ -591,7 +591,7 @@ class QueryExecutorBaseClass(metaclass=ABCMeta):
         else:
             self._on_query_completion()
 
-    def _handle_exception(self, e, stack_trace: str):
+    def _handle_exception(self, exc: Exception, stack_trace: str):
         try:
             # Try our best to fetch logs again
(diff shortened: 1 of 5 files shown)


jczhong84 (approved):

querybook/server/lib/query_executor/base_executor.py
```diff
@@ -601,11 +601,11 @@ def _handle_exception(self, e, stack_trace: str):
             pass
         finally:
             # Update logger
-            error_type, error_str, error_extracted = self._parse_exception(e)
+            error_type, error_str, error_extracted = self._parse_exception(exc)
```
jczhong84, line 604: from the [parse](https://github.com/pinterest/querybook/blob/214118c2137dfec580cba5dbeb4e0a19351f3c62/querybook/server/lib/query_executor/utils.py#L26) function, looks like this `error_type` is always `QueryExecutionErrorType.INTERNAL.value`, instead passing the `exc` to the format function, how about do some mappings for the error_type?

czgu: resolved offline

PR #958, opened 2022-08-05
feat: Ignore null values when detecting types

czgu:
For sparse columns, the type checking werent working before since there will be a lot of null values affecting the quality of the result.  

Changes include:
- sampling and null filtering is done ahead of time before feeding into detectors
- transformer only runs on not null columns
- unified null checking code

### DIFF
--- a/querybook/webapp/components/StatementResultTable/StatementResultTable.tsx
+++ b/querybook/webapp/components/StatementResultTable/StatementResultTable.tsx
@@ -5,6 +5,7 @@ import styled from 'styled-components';
 import { UserSettingsFontSizeToCSSFontSize } from 'const/font';
 import { useImmer } from 'hooks/useImmer';
 import { findColumnType } from 'lib/query-result/detector';
+import { isCellValNull } from 'lib/query-result/helper';
 import { getTransformersForType } from 'lib/query-result/transformer';
 import { IColumnTransformer } from 'lib/query-result/types';
 import { IStoreState } from 'redux/store/types';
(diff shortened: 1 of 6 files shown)


querybook/webapp/lib/query-result/helper.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * checks if a cell of results table is empty
+ * only applies to query results in querybook
+ *
+ * @param val cell from query execution result
+ * @returns boolean
+ */
+export function isCellValNull(val: any): boolean {
+    return val === 'null' || val == null;
```
jczhong84, line 9: is it possible to be empty string "" or uppercase "NULL"?

querybook/webapp/lib/query-result/detector.ts
```diff
@@ -54,6 +55,25 @@ const columnDetectors: IColumnDetector[] = [
...
+        Math.floor(rows.length / 100)
+    );
+    return columns.map((colName, index) => {
+        const notNullRowValues = rows.reduce((arr, row) => {
+            const value = row[index];
+            if (!isCellValNull(value)) {
+                arr.push(value);
+            }
+            return arr;
+        }, []);
```
jczhong84, line 70: I think the previous way + filter will be more readable, `rows.map((row) => row[index]).filter((value) => !isCellValNull(value))`
and I think it will not impact the perf much.

querybook/webapp/components/StatementResultTable/StatementResultTable.tsx
```diff
@@ -186,10 +181,11 @@ export const StatementResultTable = React.forwardRef<
                 cols={columns}
                 showPagination={showPagination}
                 formatCell={(index, column, row) => {
+                    const value = row[index];
                     const transformer = getTransformerForColumn(index);
-                    return transformer
-                        ? transformer.transform(row[index])
-                        : row[index];
+                    return transformer && !isCellValNull(value)
```
jczhong84, line 186: can we return early when `isCellValNull(value)` is true, to save the calling for `getTransformerForColumn`

querybook/webapp/lib/query-result/helper.ts
czgu: no the serialization is done by Querybook, so we know exactly what NULL would be converted to

jczhong84 (approved):

PR #960, opened 2022-08-05
Automatically add a row limit to queries

baumandm:
This change adds an optional Row Limit configuration to Query Engines.  When enabled, queries without a `LIMIT` clause will have one automatically added prior to execution, with the configured number of rows.  This is desirable to avoid inadvertently running unbounded queries on massive tables.

Users can manually disable this feature with a checkbox in the ad hoc editor, or by manually providing a `LIMIT` in their query.  This feature is not enabled for DataDocs.

Query detection support has been added for `LIMIT` and the SQL Standard `FETCH FIRST/NEXT` syntax, which covers a number of languages.  Additional parsing support would be needed to fully support other languages (e.g. SQL Server's `TOP`).

Additionally, the Trino language has been split from the Presto language to add support for [`FETCH FIRST/NEXT`](https://trino.io/blog/2020/02/03/beyond-limit-presto-meets-offset-and-ties.html), which Presto does not currently support. 

Screenshot 2022-08-04 at 11-23-56 Querybook

Screenshot 2022-08-04 at 11-24-30 Adhoc Query - Querybook

### DIFF
--- a/querybook/webapp/lib/sql-helper/sql-limiter.ts
+++ b/querybook/webapp/lib/sql-helper/sql-limiter.ts
@@ -1,0 +1,53 @@
+import {
+    containsKeyword,
+    getStatementRanges,
+    getStatementType,
+    simpleParse,
+    tokenize,
+} from './sql-lexer';
+
+/**
+ * Automatically apply a limit to a query that does not already have a limit.
+ *
+ * @param {string} query - Query to be executed.
+ * @param {rowLimit} query - Number of rows to limit the query to.
+ * @param {language} query - Language of the query.
+ * @return {string} - Query with limit applied (if necessary).
+ */
+export const getLimitedQuery = (
+    query: string,
+    rowLimit?: number,
+    language?: string
+): string => {
+    if (rowLimit == null) {
+        return query;
+    }
+
+    const statementRanges = getStatementRanges(query, null, language);
+    const statements = statementRanges
+        .map((range) => query.slice(range[0], range[1]))
+        .filter((statement) => statement.length > 0);
+
+    let addedLimit = false;
+    const updatedQuery = statements
+        .map((statement) => {
+            const tokens = tokenize(statement, { language });
+            const parsedStatements = simpleParse(tokens)[0];
+
+            if (
+                getStatementType(parsedStatements) === 'select' &&
+                !containsKeyword(parsedStatements, 'limit') &&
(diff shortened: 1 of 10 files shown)


baumandm:
auto-row-limit

querybook/webapp/lib/sql-helper/sql-limiter.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    let addedLimit = false;
+    const updatedQuery = statements
+        .map((statement) => {
+            const tokens = tokenize(statement, { language });
+            const parsedStatements = simpleParse(tokens)[0];
+
+            if (
+                getStatementType(parsedStatements) === 'select' &&
+                !containsKeyword(parsedStatements, 'limit') &&
```
czgu, line 39: this doesn't work with 
```
select * from (
   select * from table limit 5
)

```

baumandm: Good catch!  I've updated it to work with nested queries and union statements.  I also added an `(Experimental)` label to the Query Engine config to set expectations appropriately.

baumandm:
All those improvements sound good!  I'd be happy to work on adding these changes to this PR unless you would prefer to handle them.  They seem pretty straightforward so I should be able to quickly add them.

What were you thinking for DataDocs?  The ad hoc editor setting isn’t sticky, it resets if you reload to ensure someone doesn’t turn it off forever and never think about it again.  But with DataDocs, that wouldn't work very well.  I can think of two options:
1. The author must decide per-cell whether it’s on/off, and that is persistent
2. Always on with no way to turn it off

czgu:
@baumandm would it be possible to additional features in another PR  since the current code is pretty good as is 🙏 ?

The adhoc editor setting would be stored in local storage right? and in this case we can storage the cell setting in the meta field. 

I am thinking if the query engine requires a limit, we could have an option to define the range of the limit, like [1000, 10000, 100000] and there is no way to skip that limit

baumandm:
No problem! 👍 

We were thinking about this feature more like a guard rail to protect against accidents than a hard limit for the engine, which is why I went with the checkbox over a dropdown.  Would you prevent the user from manually entering a larger limit?

czgu:
i was thinking to make it like Superset's limit:
image

also to prevent users from abusing the limit, i guess we gotta think of solutions to transform the query in the backend 🤔

czgu (approved):

PR #963, opened 2022-08-08
feat: Allow executors to return warnings for query executions

ShuchiZhang:
This feature allows executors to return and display warnings. Use cases include checking different table format and warn users who are referencing inefficient format. This PR does not include any specific warning, but it can be added like:

```
class SqlAlchemyCursor(CursorBaseClass):
    _query: str = None

...

    def run(self, query):
        self._query = query
        self._cursor = self._connection.execute(sqlalchemy.sql.text(query))

...

    @property
    def warning(self):
        if self._query.__contains__("query_execution") is True:
            return "This is a boring table"
        else:
            return None
```

```
SELECT * FROM querybook2.query_execution LIMIT 30;

SELECT * FROM querybook2.query_engine;
```

Running the sql above should give the warning like:
image

### DIFF
--- a/querybook/webapp/ui/Markdown/Markdown.tsx
+++ b/querybook/webapp/ui/Markdown/Markdown.tsx
@@ -3,6 +3,7 @@ import React from 'react';
 
 import { Content } from 'ui/Content/Content';
 import { Link } from 'ui/Link/Link';
+import { Message } from 'ui/Message/Message';
 
 const MarkdownLink: React.FC<{ title: string; href: string }> = ({
     title,
@@ -19,6 +20,7 @@ const markdownOptions: MarkdownToJSX.Options = {
         a: {
             component: MarkdownLink,
         },
+        Message: { component: Message },
     },
 };
 


czgu:
hey Shuchi, thanks for adding this PR! One thing I want to point out is that query execution already has a field called `meta` which allows query engines to put arbitrary information. You can use it also to put warnings and errors as well. For warnings, you can make it force show by inserting some text as documented here https://www.querybook.org/docs/integrations/add_query_engine#meta. 

Is there still any need to add a separate warning field? I think for simplification we can look into how to display warning block with markdown used by meta

ShuchiZhang:
> hey Shuchi, thanks for adding this PR! One thing I want to point out is that query execution already has a field called `meta` which allows query engines to put arbitrary information. You can use it also to put warnings and errors as well. For warnings, you can make it force show by inserting some text as documented here https://www.querybook.org/docs/integrations/add_query_engine#meta.
> 
> Is there still any need to add a separate warning field? I think for simplification we can look into how to display warning block with markdown used by meta

Hi there, thanks for the advise. I made some changes in my implementation. I added a Markdown support for the Message. Now there are more flexibility on using the meta_info. People can display different type of messages in the meta_info section like warning message or error message. 

For example, if you want to display a warning, you can have:
```
    @property
    def meta_info(self):
        info = ""
        if self._cursor.tracking_url:
            info += f"Tracking Url: {self._cursor.tracking_url}\n"
        else:
            info += (
                '<Message type="warning" title="Test">This is a bad query!</Message>\n'
            )
            info += "---\nforce_show: true\n---"
        return info
```
The webpage will force show a warning in the meta_info section
image

czgu (approved):

PR #972, opened 2022-08-12
feat: add support of syncing table description from metastore

jczhong84:
Some systems may have the metastore as the source of truth for the table description, as well as the table schema. This change is to add the support of syncing table description from the metastore.


jczhong84:
A duplicate of https://github.com/pinterest/querybook/pull/874/

Issue #964, opened 2022-08-09
Query Engine Permissions


baumandm:
Currently, environments can be restricted to a specific set of users, but query engines associated with an environment are accessible by every user with access to the environment.

We have a use case to manage many query engines in a shared environment, but restrict each engine to specific users.  For example, one environment used by 3 teams, containing 3 query engines, each assigned to a specific team.  This is not related to visibility of queries/results, it's just about limiting who can use each engine.

I'm working on implementing this in a similar fashion to environments, with a new table `user_query_engine`.  I am planning to make query engines accessible to everyone (e.g. public) if the access control user list is empty.  If a user does not have access to a specific query engine, it would be hidden from the query engine selector.

Does this seem reasonable?  Do you have any concerns or suggestions?  Thanks!

![Screen Shot 2022-08-09 at 10 30 46 AM](https://user-images.githubusercontent.com/3084806/183675808-3307be6b-d454-4859-8cf4-bddf4bee6f13.png)

PR #975, opened 2022-08-15
feat: Adds access control to Query Engines

baumandm:
As mentioned in #964, this change adds optional access control to query engines.

Query Engines default to public accessibility, but if any users are assigned to a specific query engine, only the users in the access control list will be able to run queries on it.  If a user does not have access to a specific query engine, it will be hidden from the query engine selector.

This does not affect visibility of queries or results.


baumandm:
This PR has an issue with conflating Query Engine access and visibility.  For example, if you don't have access to an engine, ad hoc query search results show `unknown`, and data doc cells show an Invalid Query Cell warning.  I'm working on patching this.

czgu:
hey @baumandm, a quick question about this feature. Have you consider to add some authorization check on the query engine side to throw errors if the user cannot run the engine? It is much simpler to implement with plugins and requires no UI change

baumandm:
Hi @czgu, we do have authorization at the cluster-level, but it doesn't fix the UX issue of users knowing what query engines they can access.  Our environment has multiple clusters per team, along with shared clusters, so it's a bit overwhelming to show all of them at once to a user who can probably only use 3-4 of them.

czgu:
ah got it, just for the sake of brainstorming, what if we add a field called `disabled` into QueryEngine which could be set by admins that would not allow users to run query against the engine via API and UI.
Then when the user is calling `@register("/query_engine/",methods=["GET"],)`, we can add a plugin function to dynamically set the value of `disabled` 🤔 for the user

baumandm:
I don't think the plugin approach is as nice as adding API and UI support.  Being able to manage permissions for Query Engines in the UI is critical for quickly changing permissions.  We are syncing user groups from Active Directory for Environments, and we'd like to be able to do the same for Query Engines.  Latency would prevent us from being able to lookup user access on-the-fly, so we'd need to pre-calculate and cache user access per engine anyways.

However, I do think using a `disabled` field would probably simplify some of the changes needed, especially so query engines could be visible in Search/DataDocs but not executable.

czgu:
sounds good, how about lets use user_query_engine env as denylist when engine is enabled then?

baumandm:
Sorry I've been out sick so coming back to this after a few days.  The biggest issue I see with that approach is that it doesn't differentiate between query engine disabled or unauthorized, so error messages would need to be generic.  The other issue is that if I was using it as a whitelist but actually needed to disable a query engine, I'd need to remove all the users first.  So the disable toggle would be basically useless for disabling (unless you aren't using access control).

I just pushed some changes I had been working on earlier in the week to allow users to see query engines they can't access in certain locations (e.g. search results, DataDocs).  It does not use a disabled field since I started these changes before our discussion, but I could refactor it pretty easily.

The basic approach is to leave `/query_engine/` alone, and add `/user/query_engine/` that returns a list of the Query Engines that a user has access to (similar to Environments).  This is loaded into the store, and new selectors provide only accessible Query Engines.  Then specific components use either the accessible or original selectors depending on whether or not they should include all Query Engines or not.

Let me know if you prefer the disabled approach instead and I will refactor it.

baumandm:
I made a standalone PR for disabling Query Engines, #993.  It doesn't have anything related to access control, so it should be easier to review.  It borrows a lot from this PR, but doesn't add any endpoints or Redux actions so it's quite a bit simpler.

If those changes look good, I can remove most of the front-end changes from this PR and switch it to dynamically disable environments as discussed above.

PR #976, opened 2022-08-18
fix: overflow in raw metastore info

ShuchiZhang:
We notice a possible overflow in "raw metastore info" section. In the case below, some of the column names is too long to fit in one line.
image

We did some css changes to break up the string if it is too long as
image


czgu (approved):
thanks for the fix!

PR #979, opened 2022-08-24
feat: improve table upload with managed/external

czgu:
- Added plugin model for table upload exporters
- Make S3 exporter support getting path from HMS
- Added warning checks before uploading table
image

- Support uploading external or managed table

### DIFF
--- a/querybook/server/datasources/metastore.py
+++ b/querybook/server/datasources/metastore.py
@@ -99,7 +99,7 @@ def get_table_by_name(
 
 
 @register("/table_name/<schema_name>/<table_name>/exists/", methods=["GET"])
-def get_if_schema_or_table_exists(
+def get_if_schema_and_table_exists(
     metastore_id, schema_name, table_name
 ) -> Tuple[bool, bool]:
     """
--- a/querybook/server/datasources/metastore.py
+++ b/querybook/server/datasources/metastore.py
@@ -98,6 +98,38 @@ def get_table_by_name(
     return table_dict
 
 
+@register("/table_name/<schema_name>/<table_name>/exists/", methods=["GET"])
+def get_if_schema_or_table_exists(
+    metastore_id, schema_name, table_name
+) -> Tuple[bool, bool]:
+    """
+    Check if the table name / schema name exists in cache, then check the actual metastore
+    if they don't exist
+
+    Returns [schema_exists, table_exists]
+    """
+    verify_metastore_permission(metastore_id)
+    with DataTableFinder(metastore_id) as t_finder:
+        table_exists_in_cache = t_finder.get_table_by_name(schema_name, table_name)
+        if table_exists_in_cache:
+            return [True, True]
+
+        metastore_loader = get_metastore_loader(metastore_id)
+        table_exists = metastore_loader.check_if_table_exists(schema_name, table_name)
+        if table_exists:
+            return [True, True]
+
+        schema_exists_in_cache = t_finder.get_schema_by_name(schema_name)
+        if schema_exists_in_cache:
+            return [True, False]
+
+        schema_exists = metastore_loader.check_if_schema_exists(schema_name)
+        if schema_exists:
+            return [True, False]
+
+    return [False, False]
+
+
 @register("/data_job_metadata/<int:data_job_metadata_id>/", methods=["GET"])
 def get_data_job_metadata(data_job_metadata_id):
     with DBSession() as session:
--- a/querybook/server/lib/table_upload/exporter/exporter_factory.py
+++ b/querybook/server/lib/table_upload/exporter/exporter_factory.py
@@ -1,27 +1,28 @@
-from typing import Dict, List
+from typing import Dict
 
 from app.db import with_session
 from logic.admin import get_query_engine_by_id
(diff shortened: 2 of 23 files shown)


querybook/server/datasources/metastore.py
```diff
@@ -98,6 +98,38 @@ def get_table_by_name(
     return table_dict
 
 
+@register("/table_name/<schema_name>/<table_name>/exists/", methods=["GET"])
+def get_if_schema_or_table_exists(
+    metastore_id, schema_name, table_name
+) -> Tuple[bool, bool]:
+    """
+    Check if the table name / schema name exists in cache, then check the actual metastore
+    if they don't exist
```
jczhong84, line 107: is it possible that it exists in cache while it has been deleted in the actual metastore?

querybook/server/lib/table_upload/exporter/exporter_factory.py
```diff
@@ -1,27 +1,28 @@
...
 from logic.admin import get_query_engine_by_id
-from lib.utils.import_helper import import_modules
+from lib.utils.import_helper import import_module_with_default
 
 from .base_exporter import BaseTableUploadExporter
 
-ALL_TABLE_UPLOAD_EXPORTERS: List[BaseTableUploadExporter] = import_modules(
-    [
-        ("lib.table_upload.exporter.s3_exporter", "S3CSVExporter"),
-        ("lib.table_upload.exporter.s3_exporter", "S3ParquetExporter"),
```
jczhong84, line 12: why removing these 2 exporters by default?

querybook/server/lib/table_upload/exporter/s3_exporter.py
```diff
@@ -43,15 +53,56 @@ def _upload_to_s3(self) -> None:
...
+                self._table_config["schema_name"]
+            ).locationUri
+
+            # if the url is s3a or s3n, replace with s3
+            sanitized_schema_location_uri = re.sub(
+                r"^s3[a-z]:", "s3:", schema_location_uri
+            )
+
+            if not sanitized_schema_location_uri.endswith("/"):
+                sanitized_schema_location_uri += "/"
```
jczhong84, line 95: can we have a separate util function to normalize a s3 path?

Also would it be possible that the s3_path from `_export_config` contains `s3a` or `s3n`?

querybook/server/lib/table_upload/exporter/s3_exporter.py
```diff
@@ -43,15 +53,56 @@ def _upload_to_s3(self) -> None:
...
+
+            return sanitized_schema_location_uri
+
+        raise Exception("Must specify s3_path or set use_schema_location=True")
+
+    @with_session
+    def destination_s3_folder(self, session=None) -> str:
+        _, table_name = self._fq_table_name
+        object_key = table_name
+        return self.destination_s3_root(session=session) + object_key
```
jczhong84, line 105: why not just `return self.destination_s3_root(session=session) + table_name`?

querybook/webapp/components/TableUploader/TableUploaderConfirmForm.tsx
```diff
@@ -117,8 +120,99 @@ export const TableUploaderConfirmForm: React.FC = () => {
...
+                                size={20}
+                                className="mr8"
+                            />
+                            The table {schemaName}.{tableName} was detected in
+                            metastore
+                        </span>
+                    }
+                >
+                    Since the if exists policy is "fail". The table upload may
+                    not work.
```
jczhong84, line 191: since the user already selected 'fail' for the if exists policy, why would we like to show a warning instead of just failing it so they can not proceed?

querybook/server/datasources/metastore.py
czgu: yeah it is, I think this is more like a conservative check. My concern with directly with metastore is that we don't have efficient API to do so

querybook/server/lib/table_upload/exporter/exporter_factory.py
czgu: see documentation that's added, you would need to add additional configs for use them

querybook/server/lib/table_upload/exporter/s3_exporter.py
czgu: sounds good, added a shared function

as for the second point, no, because it is specified in the plugin, see documentation:
```
-   s3_path (str): if supplied, will use it as the root path for upload. Must be the full s3 path like s3://bucket/key, the trailing / is optional.
```

querybook/webapp/components/TableUploader/TableUploaderConfirmForm.tsx
czgu: hmm, i guess we could, i made it soft since the check is not in real time

querybook/server/lib/table_upload/exporter/s3_exporter.py
jczhong84: for this share function, I meant to have it also include the `s3a`, `s3n` -> `s3` replacement.

jczhong84 (approved):

PR #980, opened 2022-08-24
feat: add support of syncing table/column description from metastore

jczhong84:
Some systems may have the metastore as the source of truth for the table/column description, as well as the table schema. This change is to add the support of syncing table description from the metastore.

### DIFF
--- a/querybook/server/logic/metastore.py
+++ b/querybook/server/logic/metastore.py
@@ -271,10 +271,6 @@ def create_table_information(
         hive_metastore_description=hive_metastore_description,
     )
 
-    # As parameter descripton is newly added, for backward compatible,
-    # only set the description explicitly when it's not None.
-    # Otherwise, for those existing metastores which dont sync description
-    # to querybook, it will wipe out the existing description.
     if description is not None:
         new_table_information.description = description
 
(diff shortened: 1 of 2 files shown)


czgu (commented):
the code LGTM, a few comments:
1. can you specify the format of description that is expected? You can mention it must be HTML 
2. can you put some comments to explain the difference between description and comment field in DataColumn? it might be confusing since the naming is so similar
3. Consider adding some kind of flag in metastore loader to indicate whether or not it would provide descriptions for table and columns. That way, we can also infer this information in the frontend and disable description editing

querybook/server/logic/metastore.py
```diff
@@ -270,6 +271,13 @@ def create_table_information(
         hive_metastore_description=hive_metastore_description,
     )
 
+    # As parameter descripton is newly added, for backward compatible,
```
czgu, line 274: this comment is somewhat misleading, it is **optional** for metastore to provide description, there is no update needed for existing metastores if the admins decide not to

jczhong84: no action needed for existing metastores, just tried to explain it why we assign it separately instead of adding it directly to constructor above. 
I'll just remove this comment.

czgu (approved):

PR #983, opened 2022-08-29
DataDoc - Use Previous Query Engine

LulaV14:
We noticed there's a default query engine set that's used when creating a new query cell on a DataDoc, but we wanted to have the option of defaulting to a query cell's previous selected query engine since it's easier in the case of having multiple query engines and when users are creating query cells for the same previously used query engine.

This is the new option on the User Settings page:
image

Functionality when `Use Previous Query Engine` is enabled:
querybook-use-previous-query-engine

### DIFF
--- a/querybook/config/user_setting.yaml
+++ b/querybook/config/user_setting.yaml
@@ -20,6 +20,15 @@ default_query_engine:
     options: []
     helper: Default query engine when creating a query cell.
 
+use_previous_query_engine:
+    per_env: true
+    tab: general
+    default: disabled
+    options:
+        - enabled
+        - disabled
+    helper: Use the previous cell's selected query engine on a DataDoc when available.
+
 datadoc_arrow_key:
     default: enabled
     tab: general
--- a/querybook/webapp/components/DataDoc/DataDoc.tsx
+++ b/querybook/webapp/components/DataDoc/DataDoc.tsx
@@ -301,7 +301,8 @@ class DataDocComponent extends React.PureComponent<IProps, IState> {
         index: number,
         cellType: CELL_TYPE,
         context: string,
-        meta: IDataCellMeta
+        meta: IDataCellMeta,
+        previousEngine?: number
     ) {
         try {
             const dataDoc = this.props.dataDoc;
@@ -314,7 +315,8 @@ class DataDocComponent extends React.PureComponent<IProps, IState> {
                     index,
                     cellType,
                     context,
-                    meta
+                    meta,
+                    previousEngine
                 );
             }
         } catch (e) {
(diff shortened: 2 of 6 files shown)


querybook/webapp/components/DataDoc/DataDoc.tsx
```diff
@@ -314,7 +315,8 @@ class DataDocComponent extends React.PureComponent<IProps, IState> {
                     index,
                     cellType,
                     context,
-                    meta
+                    meta,
+                    previousEngine
```
czgu, line 319: you can bypass all of these prop passing with this:
```
const previousQueryCell = dataDoc.dataDocCells.slice(0, index).reverse().find(cell => cell.cell_type === 'query');
const previousCellEngineId = previousQueryCell?.meta?.engineId;
```

querybook/webapp/redux/dataDoc/action.ts
```diff
@@ -297,13 +298,18 @@ export function insertDataDocCell(
...
-                    ? meta['engine']
+            const defaultQueryEngine =
+                previousEngine &&
+                userSetting['use_previous_query_engine'] === 'enabled'
+                    ? previousEngine
                     : getQueryEngineId(
                           userSetting['default_query_engine'],
                           queryEngineIds
                       );
+            const engine =
```
czgu, line 309: const engine = meta?.['engine'] ?? defaultQueryEngine

czgu (commented):
thanks for adding this feature!

querybook/config/user_setting.yaml
```diff
@@ -20,6 +20,15 @@ default_query_engine:
     options: []
     helper: Default query engine when creating a query cell.
 
+use_previous_query_engine:
+    per_env: true
```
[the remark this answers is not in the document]
LulaV14: I set it per environment since the `default_query_engine` option is also set per environment I would guess it's good to have that option. But I can see how it's ok having it as a general setting as well.
Do you prefer it to be general?

querybook/webapp/components/DataDoc/DataDoc.tsx
LulaV14: Oh nice! You're right, I'll make the change

querybook/config/user_setting.yaml
```diff
@@ -20,6 +20,15 @@ default_query_engine:
...
+    tab: general
+    default: disabled
```
LulaV14, line 26: @czgu what do you feel about defaulting this feature to `enabled` instead?

czgu (commented):
code lgtm, will r+ once the user settings change is done 👍

querybook/config/user_setting.yaml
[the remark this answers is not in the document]
LulaV14: Sounds good 👍

czgu (approved):

czgu:
@LulaV14 before merge, can you fix the lint error? just for future reference, you can use pre-commit https://www.querybook.org/docs/developer_guide/developer_setup to do it automatically

LulaV14:
> @LulaV14 before merge, can you fix the lint error? just for future reference, you can use pre-commit https://www.querybook.org/docs/developer_guide/developer_setup to do it automatically

Thanks for pointing that out, I was using Github Codespaces and forgot to install the pre-commit hook 😅

czgu (approved):

PR #984, opened 2022-08-29
feat: add query validation to querybook

czgu:
added live syntax check to querybook.
The only default support right now is Presto/Trino.

Example UI:
image
image

Multi query support:
image


querybook/webapp/lib/templated-query/validation.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * This is a super basic function to check whether or not a string
+ * uses jinja templating, it should only be used as a safety measure
+ * since it does not provide perfect correctness
+ *
+ * @param query
+ * @returns boolean
+ */
+export function isQueryUsingTemplating(query: string): boolean {
```
jczhong84, line 9: so it wont cover the case of template strings are in the query comments?

querybook/webapp/components/QueryEditor/QueryEditor.tsx
```diff
@@ -183,27 +188,47 @@ export class QueryEditor extends React.PureComponent<
...
-                                delay: 1000,
-                            }
-                          : null,
-                  }
-                : {};
+        const lintingOptions = getLintErrors
+            ? {
+                  lint: {
+                      // Lint only when you can edit
+                      getAnnotations: async (
```
jczhong84, line 204: do you wanna have a separate function for it?  it seems too complicated as an inline one to me.

querybook/server/lib/query_analysis/validation/validators/presto_explain_validator.py
```diff
@@ -0,0 +1,109 @@
...
+            statement_error_ch,
+            QueryValidationSeverity.ERROR,
+            error_msg,
+        )
+
+    def _run_validation_statement(self, statement: str, engine_id: int, uid: int):
+        ExecuteQuery(False, 0.1)(statement, engine_id, uid)
+
+    def language(self):
+        return "presto"
```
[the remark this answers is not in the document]
czgu: hm i guess not really. We could add it as a refactor PR since there are a lot of places to change

querybook/webapp/lib/templated-query/validation.ts
czgu: tetmplate strings in query comments are still rendered, and if it is multi line, it would impact the query still, so there is no good way to prevent that

jczhong84 (approved):

jczhong84:
Please fix the nodetests before merging

PR #992, opened 2022-09-02
feat: pass execution type to executor client

jczhong84:
Add query execution type as enum, and pass it down to the executor client
* adhoc
* scheduled

with that, we can pass the scheduled flag to the Presto client, something like 
```
connection = presto.connect(
...
session_props={'presto_scheduled_query': 'true'}, # when execution type is scheduled
)

```

### DIFF
--- a/querybook/server/datasources/datadoc.py
+++ b/querybook/server/datasources/datadoc.py
@@ -10,7 +10,6 @@ from app.datasource import register, api_assert, with_impression
 from app.flask_app import socketio
 from app.db import DBSession, with_session
 from const.impression import ImpressionItemType
-from const.query_execution import QueryExecutionType
 from env import QuerybookSettings
 
 from lib.celery.cron import validate_cron
@@ -291,8 +290,6 @@ def create_datadoc_schedule(
                 **kwargs,
                 "user_id": current_user.id,
                 "doc_id": id,
-                "scheduled": True,
-                "execution_type": QueryExecutionType.SCHEDULED.value,
             },
             task_type="user",
             session=session,
--- a/querybook/server/datasources/datadoc.py
+++ b/querybook/server/datasources/datadoc.py
@@ -10,6 +10,7 @@ from app.datasource import register, api_assert, with_impression
 from app.flask_app import socketio
 from app.db import DBSession, with_session
 from const.impression import ImpressionItemType
+from const.query_execution import QueryExecutionType
 from env import QuerybookSettings
 
 from lib.celery.cron import validate_cron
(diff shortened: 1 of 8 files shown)


querybook/server/datasources/datadoc.py
```diff
@@ -286,7 +287,13 @@ def create_datadoc_schedule(
             schedule_name,
             "tasks.run_datadoc.run_datadoc",
             cron=cron,
-            kwargs={**kwargs, "user_id": current_user.id, "doc_id": id},
+            kwargs={
+                **kwargs,
+                "user_id": current_user.id,
+                "doc_id": id,
+                "scheduled": True,
```
czgu, line 294: this parameter doesn't do anything

querybook/server/datasources/datadoc.py
```diff
@@ -286,7 +287,13 @@ def create_datadoc_schedule(
...
+                "execution_type": QueryExecutionType.SCHEDULED.value,
```
czgu, line 295: since its defaulted to `QueryExecutionType.SCHEDULED.value`, don't put the value here?

querybook/server/lib/query_executor/executor_factory.py
```diff
@@ -36,6 +43,7 @@ def _get_executor_params_and_engine(query_execution_id, celery_task, session=Non
         raise ArchivedQueryEngine("This query engine is disabled.")
 
     client_setting = get_client_setting_from_engine(engine, uid, session=session)
+    client_setting["execution_type"] = execution_type
```
czgu, line 46: i don't think ClientBaseClass needs to know `execution_type`, it is more of a adapter for different dbapi. execution_type can be passed to QueryExecutor and it can determine what flags to pass to query client

querybook/server/datasources/datadoc.py
jczhong84: forgot to remove

querybook/server/lib/query_executor/executor_factory.py
jczhong84: this `client_setting` will pass to `QueryExecutorBaseClass._client_setting`, and it will use it to create query client
```    
def _get_cursor(self):
        if self._client is None:
            self._client = self._get_client(self._client_setting)

        return self._client.cursor()
```

or are you talking another path to pass it down?

querybook/server/datasources/datadoc.py
jczhong84: I was thinking we may support to adhocly run an entire datadoc in the future. but you're right, we could just pass `QueryExecutionType.ADHOC` for it and keep the default value in `run_datadoc` task as `scheduled`

querybook/webapp/const/queryExecution.ts
```diff
@@ -1,6 +1,11 @@
 import { ICancelablePromise } from 'lib/datasource';
 
 // Keep this the same as the Enum defined in const/query_execution.py
```
czgu, line 3: since the line didn't get moved, can you change it to 

```
/*
   Keep definitions in this file to be same as const/query_execution.py
*/
```

czgu (approved):

czgu (approved):

PR #993, opened 2022-09-06
feat: Allow disabling Query Engines

baumandm:
This feature allows Query Engines to be disabled without deleting or removing them from an environment.  Disabled Query Engines cannot be used to execute queries and they do not appear in the Query Engine dropdown.  However, they do appear in search results, search filters, and where used in DataDocs and snippets.

Screen Shot 2022-09-06 at 6 21 09 PM

### DIFF
--- a/querybook/webapp/redux/queryEngine/selector.ts
+++ b/querybook/webapp/redux/queryEngine/selector.ts
@@ -33,8 +33,25 @@ export const queryEngineSelector = createSelector(
     (queryEngineById, engineIds) => engineIds.map((id) => queryEngineById[id])
 );
 
+export const enabledQueryEngineSelector = createSelector(
+    queryEngineSelector,
+    (queryEngines) =>
+        queryEngines.filter(
+            (engine) => engine.feature_params?.disabled !== true
+        )
+);
+
 export const queryEngineByIdEnvSelector = createSelector(
     queryEngineSelector,
+    (queryEngines) =>
+        new Proxy(arrayGroupByField(queryEngines), {
+            get: (engineById, name: string) =>
+                name in engineById ? engineById[name] : unknownQueryEngine,
+        })
+);
+
+export const enabledQueryEngineByIdEnvSelector = createSelector(
+    enabledQueryEngineSelector,
     (queryEngines) =>
         new Proxy(arrayGroupByField(queryEngines), {
             get: (engineById, name: string) =>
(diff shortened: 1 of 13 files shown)


querybook/webapp/redux/queryEngine/selector.ts
```diff
@@ -42,16 +50,32 @@ export const queryEngineByIdEnvSelector = createSelector(
         })
 );
 
+export const enabledQueryEngineByIdEnvSelector = createSelector(
```
czgu, line 53: is this used anywhere?

querybook/webapp/redux/queryEngine/selector.ts
```diff
@@ -42,16 +50,32 @@ export const queryEngineByIdEnvSelector = createSelector(
...
     queryEngineStatusByIdSelector,
     engineIdsInEnvironmentSelector,
-    (queryEngineStatusById, engineIds) =>
-        engineIds.reduce((pairs, id) => {
-            if (id in queryEngineStatusById) {
-                pairs.push([id, queryEngineStatusById[id]]);
-            }
-            return pairs;
-        }, []) as Array<[number, IQueryEngineStatus]>
+    enabledQueryEngineSelector,
```
czgu, line 65: update fetchAllSystemStatus to not fetch disabled engines, then the filter would be no longer needed?

querybook/webapp/redux/queryEngine/selector.ts
```diff
@@ -42,16 +50,32 @@ export const queryEngineByIdEnvSelector = createSelector(
...
+    (queryEngineStatusById, engineIds, enabledQueryEngines) =>
+        engineIds
+            .filter(
```
czgu, line 68: since there is a filtering layer, we could write this as

```
.filter(id => enabledQueryEngines.some((engine) => engine.id === id) && id in queryEngineStatusById)
.map(id => [id, queryEngineStatusById[id]])
```

querybook/webapp/redux/queryEngine/selector.ts
baumandm: Good catch, I forgot to update that function.  But this filter is still needed to avoid showing all clusters in the status popover.

baumandm:
Hey @czgu could you take another look at this when you get a chance?  Thanks!

PR #998, opened 2022-09-12
feat: Add a new function to sync table with metastore

jczhong84:
For syncing a table with metastore, the caller may not know if it's an updating or deleting, it should be the sync function's responsibility to keep it sync with metastore, whether it's creating, updating or deleting.

Also the `get_all_table_names_in_schema` method used for checking table existence is actually not very efficient for those schema with a huge amount of tables, it could cost 1 to 2 seconds. So this new function will use `get_table_and_columns` to check, and its response will also be used for creating/updating the table as it will be needed anyway if it's not deleting.

### DIFF
--- a/querybook/server/datasources/metastore.py
+++ b/querybook/server/datasources/metastore.py
@@ -671,25 +671,18 @@ def get_schemas(metastore_id, limit=5, offset=0, sort_key="name", sort_order="de
     "/table/<schema_name>/<table_name>/sync/",
     methods=["PUT"],
 )
-def sync_table_from_metastore(
-    schema_name, table_name, metastore_id, is_delete: bool = False
-):
-    """Sync table info from metastore. Delete the table if is_delete is True.
+def sync_table_from_metastore(schema_name, table_name, metastore_id):
+    """Sync table info with metastore.
 
     Args:
-        metastore_id (int): metastore ID
+        metastore_id (int): Metastore ID
         schema_name (str): Schema name
         table_name (str): Table name
 
     Returns:
-        None if deleting a table
+        None if the table doesn't exist neither metastore nor querybook
+        -1 if table is deleted
         table id if creating or updating a table
     """
     metastore_loader = get_metastore_loader(metastore_id)
-
-    if is_delete:
-        metastore_loader.sync_delete_table(schema_name, table_name)
-        return None
-    else:
-        table_id = metastore_loader.sync_create_or_update_table(schema_name, table_name)
-        return table_id
+    return metastore_loader.sync_table(schema_name, table_name)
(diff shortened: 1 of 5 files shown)


querybook/server/tasks/log_query_per_table.py
```diff
@@ -80,46 +80,35 @@ def create_lineage_from_query(
...
+    tables_to_sync = set()
     for tables, statement_type in zip(table_per_statement, statement_types):
-        if statement_type == "DROP":
+        if statement_type in ("CREATE", "ALTER", "DROP"):
             for table in tables:
-                tables_to_add.discard(table)
-                tables_to_remove.add(table)
-        elif statement_type is not None:  # Any other DML/DDL
+                tables_to_sync.add(table)
+        elif statement_type is not None:
```
czgu, line 97: keep the comments since they are still applicable
```
# Otherwise for things like insert/select we only update
# if it doesn't exist in the metastore

File: querybook/server/tasks/log_query_per_table.py [hunk 2]
```diff
@@ -80,46 +80,35 @@ def create_lineage_from_query(
...
-                        )
-                        if not query_table:
-                            tables_to_add.add(table)
-
-    for table in tables_to_remove:
-        schema_name, table_name = table.split(".")
-        metastore_loader.sync_delete_table(schema_name, table_name, session=session)
-
-    for table in tables_to_add:
+                schema_name, table_name = table.split(".")
```
czgu, line 99: still need 
```
if table not in tables_to_add
```
to reduce the amount of checks

querybook/server/datasources/metastore.py
```diff
@@ -671,25 +669,18 @@ def get_schemas(metastore_id, limit=5, offset=0, sort_key="name", sort_order="de
     "/table/<schema_name>/<table_name>/sync/",
     methods=["PUT"],
 )
-def sync_table_from_metastore(
-    schema_name, table_name, metastore_id, is_delete: bool = False
-):
-    """Sync table info from metastore. Delete the table if is_delete is True.
+def sync_table_from_metastore(schema_name, table_name, metastore_id):
```
czgu, line 672: nit: since the code between refresh and sync are so similar, can you update the name to be more consistent?

like
`sync_table_by_table_id`
`sync_table_by_name`

querybook/server/datasources/metastore.py
```diff
@@ -214,9 +214,7 @@ def refresh_table_from_metastore(table_id):
 
         metastore_id = schema.metastore_id
         metastore_loader = get_metastore_loader(metastore_id, session=session)
-        metastore_loader.sync_create_or_update_table(
-            schema.name, table.name, session=session
-        )
+        metastore_loader.sync_table(schema.name, table.name, session=session)
```
czgu, line 217: what happens if table is deleted, and we call `session.refresh(table)`?

also the frontend code assumes a table is always returned, can you add some checks on the frontend too? (like remove it from redux store, show messages, etc)

czgu:
Let's address the follow ups in another PR, to handle edge cases like refreshing a deleted table

czgu (approved):

PR #985, opened 2022-08-30
Bugfix: scatter/bubble charts were losing their labels after save

pateketrueke:
Hi, I was asked to investigate an issue involving the `datalabels` plugin, as it was rendering `[object Object]` instead of any other value... after inspecting the integration code I realized that `chartData['labels']` is not always populated!

Then, I tried to play with the `datalabels.formatter` callback to see if was possible to get the missing label from there but it was not possible, after you save the chart options then `context.chart.data.labels` becomes an empty array.

It seems whenever the `chartjs-plugin-datalabels` receives an `undefined` value from any formatter then it renders `[object Object]` instead? &mdash; my workaround sets `charData['labels']` from the first-field on `dataRows` because I am not using `x_axis` (`xAxisIdx` is not zero!) as the same value for the labels, e.g.

1. First, we create some sample data:

```sql
CREATE TABLE test_table(
   id   INTEGER NOT NULL PRIMARY KEY
  ,name VARCHAR(41) NOT NULL
  ,a    INTEGER  NOT NULL
  ,b    INTEGER  NOT NULL
  ,c    INTEGER  NOT NULL
  ,d    INTEGER  NOT NULL
);
INSERT INTO test_table(name,a,b,c,d) VALUES ('extinct',2278286,14347746,99079961,1927908);
INSERT INTO test_table(name,a,b,c,d) VALUES ('disappear',737184,30059,31556,12812);
INSERT INTO test_table(name,a,b,c,d) VALUES ('horizon',79358,404477,61049325,24140);
INSERT INTO test_table(name,a,b,c,d) VALUES ('command',7692138,4909601,83163263,418289);
INSERT INTO test_table(name,a,b,c,d) VALUES ('visual',624265,14360365,99057338,1905921);
INSERT INTO test_table(name,a,b,c,d) VALUES ('neighborhood',734783,30059,31555,12811);
INSERT INTO test_table(name,a,b,c,d) VALUES ('modernize',37830126,0,436967,432945);
INSERT INTO test_table(name,a,b,c,d) VALUES ('complain',37855786,462964,36796556,1136975);
INSERT INTO test_table(name,a,b,c,d) VALUES ('retired',37842292,677455,61470550,1730112);
INSERT INTO test_table(name,a,b,c,d) VALUES ('tell',6356376,54380358,57275104,1460489);
INSERT INTO test_table(name,a,b,c,d) VALUES ('deprive',4078907,0,42266,42026);
[...]

PR #999, opened 2022-09-12
feat: improve charting axis and value display

czgu:
- Update chart.js to latest
- Added option to format axis with either $ or %
- also semi reverted the changes in https://github.com/pinterest/querybook/pull/985. Instead of using the first column as value (for scatter). Added option to either use "series" name or y value for label
- fix: query transpilation doesn't save to backend
- updated run button length (back to 120px), shrunk the ... button


querybook/webapp/components/QueryRunButton/QueryRunButton.tsx
```diff
@@ -215,7 +215,7 @@ const QueryLimitSelector: React.FC<{
             customButtonRenderer={() => (
                 <div
                     className="flex-center ph4"
-                    aria-label="Only applies to select statements without limit"
+                    aria-label="Only applies to SELECT query without LIMIT"
```
jczhong84, line 218: do we want to provide the option of disabling the limit feature, even the query doesn't have `LIMIT`?

czgu: sounds good, in another PR? (since this one is getting a bit messy now)

jczhong84 (approved):

PR #1000, opened 2022-09-13
feat: allow user to bypass LIMIT

czgu:
Added a new flag ALLOW_UNLIMITED_QUERY which is turned on by default. This means there is an new option "none" to skip the limit query postprocessing

image

If user did pick none and ran a query without limit, they are shown this warning modal every time:
image

### DIFF
--- a/querybook/webapp/components/QueryComposer/RunQuery.tsx
+++ b/querybook/webapp/components/QueryComposer/RunQuery.tsx
@@ -106,13 +106,14 @@ async function transformLimitedQuery(
             message: (
                 <Content>
                     <div>
-                        The following SELECT statement have no limit. Please
-                        make sure you intend to get all the rows returned.
+                        The following SELECT statement has no limit. Please make
+                        sure you intend to get all the rows returned.
                     </div>
                     <div>
                         <i>
                             Tip: to avoid seeing this message, add a LIMIT in
-                            the query or set it on left of the run button.
+                            the query or set a limit value on the left of the
+                            run button.
                         </i>
                     </div>
                     <pre>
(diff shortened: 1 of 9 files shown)


querybook/webapp/components/QueryComposer/RunQuery.tsx
```diff
@@ -71,14 +75,58 @@ async function transformTemplatedQuery(
...
+            message: (
+                <Content>
+                    <div>
+                        The following SELECT statement have no limit. Please
+                        make sure you intend to get all the rows returned.
+                    </div>
+                    <div>
+                        <i>
+                            Tip: to avoid seeing this message, add a LIMIT in
+                            the query or set it on left of the run button.
```
jczhong84, line 115: not good at English grammar, but feel like it should be `on the left` :) 

or maybe we can say `or pick a limit from the Limit dropdown list`

querybook/webapp/lib/sql-helper/sql-lexer.ts
```diff
@@ -573,6 +573,24 @@ export function getStatementRanges(
     return statementRanges;
 }
 
+/**
+ * Split a query into multiple statements. Return the non-emptry ones
```
jczhong84, line 577: typo: `non-empty`

querybook/webapp/components/QueryRunButton/QueryRunButton.tsx
```diff
@@ -204,6 +208,13 @@ const QueryLimitSelector: React.FC<{
         }
     }, [rowLimit, setRowLimit]);
 
+    const selectedRowLimitText = React.useMemo(() => {
+        if (rowLimit != null && rowLimit >= 0) {
```
jczhong84, line 212: as the `rowLimit` has type of `number`, why need to check if it's null? can't it be simply 
`rowLimit === -1 ? 'none' : formatNumber(rowLimit)`?

querybook/webapp/lib/sql-helper/sql-limiter.ts
```diff
@@ -67,6 +67,24 @@ export function getSelectStatementLimit(
     return -1;
 }
 
+/**
+ * Check if the query have any statements that is SELECT and does not have LIMT
```
jczhong84, line 71: `query have` -> `query has`

typo: `LIMT` -> `LIMIT`

jczhong84 (approved):

PR #1003, opened 2022-09-13
feat: show 404 page when table gets deleted

jczhong84:
When clicking "Refresh from metastore", if the table gets deleted, we'll show an empty 404 page.
sync-deleted

Also the previous error page of table not found is not rendered properly.

**Before**
error_modal
error_page

**After**
404_page

### DIFF
--- a/querybook/server/datasources/metastore.py
+++ b/querybook/server/datasources/metastore.py
@@ -72,7 +72,11 @@ def get_table(table_id, with_schema=True, with_column=True, with_warnings=True):
     # username = flask_session['uid']
     with DBSession() as session:
         table = logic.get_table_by_id(table_id, session=session)
-        api_assert(table, "Invalid table")
+        api_assert(
+            table,
+            "Table doesn't exist or has been deleted from Metastore",
+            status_code=404,
+        )
         verify_data_schema_permission(table.schema_id, session=session)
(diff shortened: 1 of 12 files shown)


querybook/webapp/hooks/useMounted.ts
```diff
@@ -1,12 +1,15 @@
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useRef } from 'react';
 
 export function useMounted() {
```
jczhong84, line 3: This old `useMounted` hook doesn't work actually.  There is good reading about it https://www.codemzy.com/blog/ismounted-hook-with-useeffect-reactjs

querybook/server/datasources/metastore.py
```diff
@@ -214,7 +221,9 @@ def refresh_table_from_metastore(table_id):
 
         metastore_id = schema.metastore_id
         metastore_loader = get_metastore_loader(metastore_id, session=session)
-        metastore_loader.sync_table(schema.name, table.name, session=session)
+        table_id = metastore_loader.sync_table(schema.name, table.name, session=session)
+        if table_id == -1:
+            return -1
```
czgu, line 226: return None? since returning a Union[int, Dict] is a bit confusing for typing

czgu (approved):

PR #1013, opened 2022-09-20
fix: url validation

czgu:
changed to regex since the previous one would capture json for some reason
also added some unit tests to show its capability

### DIFF
--- a/querybook/webapp/lib/utils/index.ts
+++ b/querybook/webapp/lib/utils/index.ts
@@ -296,9 +296,9 @@ export function getChangedObject(
     return ret;
 }
 
-// Source: https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url/45567717#45567717
+// Source: https://stackoverflow.com/questions/161738/what-is-the-best-regular-expression-to-check-if-a-string-is-a-valid-url
 const urlPattern =
-    /(?:https?):\/\/(\w+:?\w*)?(\S+)(:\d+)?(\/|\/([\w#!:.?+=&%!\-\/]))?/;
+    /^(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,5}(:[0-9]{1,5})?(\/.*)?$/i;
 export function isValidUrl(url: string): boolean {
     return !!urlPattern.test(url);
 }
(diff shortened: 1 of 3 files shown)


querybook/webapp/ui/Link/Link.tsx
```diff
@@ -27,6 +27,17 @@ function isInternalUrl(url: LocationDescriptor): boolean {
     return false;
 }
 
+function getExternalUrl(url: LocationDescriptor) {
```
jczhong84, line 30: for external url, whether it starts with `https?` or not, I think it's fine to put it in the `href`.  like `<a href="google.com">google</a>`

or are you trying to solve any other issue?

querybook/webapp/lib/utils/index.ts
```diff
@@ -296,9 +296,9 @@ export function getChangedObject(
     return ret;
 }
 
-// Source: https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url/45567717#45567717
+// Source: https://stackoverflow.com/questions/161738/what-is-the-best-regular-expression-to-check-if-a-string-is-a-valid-url
 const urlPattern =
-    /(?:https?):\/\/(\w+:?\w*)?(\S+)(:\d+)?(\/|\/([\w#!:.?+=&%!\-\/]))?/;
+    /^(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[a-z0-9]+([\-\.]{1}[a-z0-9]+)*\.[a-z]{2,5}(:[0-9]{1,5})?(\/.*)?$/i;
```
jczhong84, line 301: we may simplify `(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?` as `(https?:\/\/)?`, which can also pass all the tests

querybook/webapp/ui/Link/Link.tsx
czgu: that example you put would not redirect to a new site, https://stackoverflow.com/questions/43803778/href-without-https-prefix

querybook/webapp/lib/utils/index.ts
czgu: good point

querybook/webapp/ui/Link/Link.tsx
jczhong84: you're right. seems prepending `//` is a better option. also can we merge the check as 
```
if (url && typeof url === 'string' && !/^[a-zA-Z]+:\/\//.test(url)) {
  return `//${url}`
}

return url
```

jczhong84 (approved):

PR #1014, opened 2022-09-22
feat: show deactivated user with different ui

czgu:
Disabled users are now searchable in user search.
They will also look different in UI, with grayed out icon and `(deactivated)` added to their name:
image

Also added a task to disable scheduled docs if the owner is deleted, this task is optional

### DIFF
--- a/querybook/server/logic/schedule.py
+++ b/querybook/server/logic/schedule.py
@@ -182,7 +182,7 @@ def get_scheduled_data_docs_by_user(
         session.query(DataDoc, TaskSchedule)
         .join(
             TaskSchedule,
-            TaskSchedule.name == func.concat(DATADOC_SCHEDULE_PREFIX, DataDoc.id),
+            TaskSchedule.name == get_data_doc_schedule_name(DATADOC_SCHEDULE_PREFIX),
             isouter=(not filters.get("scheduled_only", False)),
         )
         .filter(DataDoc.owner_uid == uid)
--- a/querybook/server/logic/schedule.py
+++ b/querybook/server/logic/schedule.py
@@ -182,7 +182,7 @@ def get_scheduled_data_docs_by_user(
         session.query(DataDoc, TaskSchedule)
         .join(
             TaskSchedule,
-            TaskSchedule.name == get_data_doc_schedule_name(DATADOC_SCHEDULE_PREFIX),
+            TaskSchedule.name == func.concat(DATADOC_SCHEDULE_PREFIX, DataDoc.id),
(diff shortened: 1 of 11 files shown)


querybook/server/logic/schedule.py
```diff
@@ -180,7 +182,7 @@ def get_scheduled_data_docs_by_user(
         session.query(DataDoc, TaskSchedule)
         .join(
             TaskSchedule,
-            TaskSchedule.name == func.concat("run_data_doc_", DataDoc.id),
+            TaskSchedule.name == func.concat(DATADOC_SCHEDULE_PREFIX, DataDoc.id),
```
jczhong84, line 185: why not just use `get_data_doc_schedule_name` ^

querybook/webapp/components/DataDocViewersBadge/DataDocViewersBadge.tsx
```diff
@@ -131,7 +131,12 @@ export const DataDocViewersBadge = React.memo<IDataDocViewersBadgeProps>(
                             uid: viewerInfo.uid,
                             tooltip:
                                 viewerInfo.uid in userInfoById
-                                    ? userInfoById[viewerInfo.uid].username
+                                    ? (userInfoById[viewerInfo.uid].fullname ??
+                                          userInfoById[viewerInfo.uid]
+                                              .username) +
+                                      (userInfoById[viewerInfo.uid].deleted
+                                          ? ' (deactivated)'
+                                          : '')
```
jczhong84, line 139: can we have separate function for it? also it's not very readable as it has nested ternary operations to me.

btw for the elastics search, you have `"fullname": (fullname or username) + (" (deactivated)" if user.deleted else "")` on the backend, while this one is handled on frontend, and also seeing something similar below like `@{userName} {deletedText}` (UserBadge.tsx)

Would it be better to make the fullname transform either on the backend and can have a shared function to transformation, or it should be on the frontend, so the string of ` (deactivated)` could be more customizable?

czgu: there are 2 sources of data:
- db
- elasticsearch
so i am adding `(deactivated)` for data from db. And for elasticsearch i am adding it to the full name field. Ideally, we should add a `deleted` field in elasticsearch so we don't have to do any transformation in the backend. However, i think we can do that later since it is somewhat of a breaking change

querybook/server/logic/schedule.py
czgu: DataDoc.id is not a string, its a placeholder variable in SQL

PR #1021, opened 2022-09-27
fix: do not transform presto map keys

czgu:
This is a fix for #991 
example query that fails:
```sql
select MAP(ARRAY[ARRAY['foo' ], ARRAY['bar']], ARRAY[CAST(ROW(1, 2.0) AS ROW(x BIGINT, y DOUBLE)), CAST(ROW(3, 4.0) AS ROW(x BIGINT, y DOUBLE))])
```
The reason is that MAP is converted to Python dict, but Python dict cannot have keys with type `list` (which is what ARRAY gets translated to).

When Presto output rows in JSON, the keys are already serialized to strings, which means they are not parsable anyway, in this case, we just do not translate the keys at all.

Also added testing


czgu:
@froukees FYI, let us know if you have alternative suggestions

jczhong84 (approved):

PR #1027, opened 2022-10-03
fix: schedule owner not update

jczhong84:
with operator `|=`,  the id of `task_schedule.kwargs` remains the same, in which case `SQLAlchemy` will treat it as not mutated.

### DIFF
--- a/querybook/server/logic/schedule.py
+++ b/querybook/server/logic/schedule.py
@@ -116,7 +116,7 @@ def update_datadoc_schedule_owner(
     if not task_schedule:
         return
 
-    task_schedule.kwargs |= {"user_id": owner_id}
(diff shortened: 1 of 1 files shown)


querybook/server/logic/schedule.py
```diff
@@ -116,7 +116,7 @@ def update_datadoc_schedule_owner(
     if not task_schedule:
         return
 
-    task_schedule.kwargs |= {"user_id": owner_id}
+    task_schedule.kwargs = {**task_schedule.kwargs, "user_id": owner_id}
```
czgu, line 119: i guess sqlalchemy doesn't understand the |= operator yet

czgu (approved):
nice

PR #1028, opened 2022-10-04
fix: Replace cronjs-matcher with cron-parser

### LINKED ISSUE #997: Next Run on Schedules page is incorrect for certain cron expressions
We've noticed that the Next Run feature on the Schedules page displays incorrectly for certain schedules, specifically weekly and monthly schedules. They show today (or tomorrow depending on time) instead of the actual future date.

As far as I can tell, this is entirely cosmetic, and the schedules execute correctly.

Screen Shot 2022-09-08 at 4 20 11 PM

This shows Sept 8 (the date of the screenshot), but it should have been Sept 12 (the subsequent Monday). 
Screen Shot 2022-09-08 at 4 20 25 PM

I dug into the code and it looks to be an issue specifically in [`@datasert/cronjs-matcher`](https://github.com/datasert/cronjs).  I made a simple Code Sandbox that reproduces the issue: https://codesandbox.io/s/happy-parm-c85z3m.  

This dependency is only used for this feature, and it appears to be abandoned as the latest version was never published to npm.  I would suggest replacing it with a more popular package like [cron-parser](https://www.npmjs.com/package/cron-parser).

I can work up a PR for this if that sounds good.

baumandm:
Fixes #997 by replacing `@datasert/cronjs-matcher` with `cron-parser`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "querybook",
-    "version": "3.11.2",
+    "version": "3.11.3",
     "description": "A Big Data Webapp",
     "private": true,
     "scripts": {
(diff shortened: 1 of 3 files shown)


czgu (commented):
thanks for the fix! @baumandm can you update the version in package.json before merge?

czgu (approved):

PR #1030, opened 2022-10-05
fix: Increase CodeMirror hint z-index above modals

baumandm:
The autocomplete feature in the Query editor currently has a z-index of 30, whereas modals are 300.  This causes issues on the Snippet editor:

Screen Shot 2022-10-05 at 9 20 12 AM

This bumps the z-index up to 400 to fix:

Screen Shot 2022-10-05 at 9 19 47 AM


czgu (approved):
awesome fix!

PR #1038, opened 2022-10-12
feat: update base and demo dag exporter

jczhong84:
some updates to the dag exporter
* change the exporter type `text` to `code`
* update the returned export data with some additional fields like `name`, `success`, `msg`, `url` and `code`
* add a new modal component to render the returned export data

image

### DIFF
--- a/querybook/webapp/hooks/dag/useExporterSettings.ts
+++ b/querybook/webapp/hooks/dag/useExporterSettings.ts
@@ -23,7 +23,7 @@ function useExporterDataByName() {
         if (exporterNames.length === 0) {
             dispatch(fetchDAGExporters());
         }
-    }, [dispatch]);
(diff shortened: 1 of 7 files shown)


querybook/webapp/hooks/dag/useExporterSettings.ts
```diff
@@ -23,7 +23,7 @@ function useExporterDataByName() {
         if (exporterNames.length === 0) {
             dispatch(fetchDAGExporters());
         }
-    }, [dispatch]);
+    }, [dispatch, exporterNames.length]);
```
czgu, line 26: if there is no DAG exporter provided, this would infinite loop?

when `dispatch(receiveDAGExporters(exporters))`, we can probably not dispatch if there are no exporters

czgu (approved):

PR #1043, opened 2022-10-14
ui: keep the search string when switching the search type

jczhong84:
When switching the search type tab, we'd like to keep the search string the same, to make it the behavior consistent with popular search engines.
image

Also updated some the date filters for the table and datadoc tab as `Created At` to make it less confused.

### DIFF
--- a/querybook/webapp/redux/search/reducer.ts
+++ b/querybook/webapp/redux/search/reducer.ts
@@ -114,11 +114,10 @@ export default function search(
                     searchFilters: draft.searchFilters,
                 };
 
-                // rehydrate from past state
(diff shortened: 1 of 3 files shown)


querybook/webapp/redux/search/reducer.ts
```diff
@@ -131,7 +130,6 @@ export default function search(
                 };
 
                 draft.searchType = action.payload.searchType;
-                draft.searchString = searchStateForNewType.searchString;
```
czgu, line 134: can you also delete the type searchString from pastSearchStateByType

czgu (approved):

PR #1046, opened 2022-10-21
fix: template variable type change and big numbers

jczhong84:
**Issue**
1. when changing the var type from string to number or vice versa, the type will actually not get updated after refreshing the page.
2. for number type with really big values, it will be rounded

**Cause**
1. when only changing the `valueType`, but not update the value field, the value field will not get re-rendered so that the value will not be re-evaluated. So when clicking save, it will still have the old type.
2. The maximum number Javascript supports is `Number.MAX_SAFE_INTEGER` (2^53-1 )

**Fix**
* evaluate the value according to the type when saving
* update the `Yup` shcema validation for numbers bigger than `Number.MAX_SAFE_INTEGER`  to ask user to use string type instead
* use `SimpleField` with `type={'input'}` for both `string` and `number` type. 
  * The reason of this change is currently `number` type is using `NumberInput` and it will do a conversion from string to number `onChange(value === '' ? null : Number(value));` , which will result in the rounding issue when user input a big number first and then change the type to string. 

Snip20221020_49

### DIFF
--- a/querybook/webapp/components/DataDocTemplateButton/DataDocTemplateVarForm.tsx
+++ b/querybook/webapp/components/DataDocTemplateButton/DataDocTemplateVarForm.tsx
@@ -8,6 +8,7 @@ import { IconButton } from 'ui/Button/IconButton';
 import { SimpleField } from 'ui/FormikField/SimpleField';
 
 import {
+    getVariableValueByType,
     detectVariableType,
     SUPPORTED_TYPES,
     TTemplateVariableDict,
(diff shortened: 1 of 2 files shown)


querybook/webapp/components/DataDocTemplateButton/DataDocTemplateVarForm.tsx
```diff
@@ -46,20 +61,36 @@ export const DataDocTemplateVarForm: React.FunctionComponent<
...
+                        name: key,
+                        valueType: detectVariableType(value),
+                        value,
+                    } as const)
             ),
         }),
         [defaultTemplatedVariables, templatedVariables]
     );
+
+    const getValueByType = (value, valueType) => {
```
czgu, line 74: valueType: TSupportedTypes

czgu (approved):

PR #1050, opened 2022-10-21
ui: update dag exporter

meowcodes:
Screen Shot 2022-10-21 at 2 57 59 PM
Screen Shot 2022-10-21 at 3 01 00 PM
Screen Shot 2022-10-21 at 3 07 56 PM

### DIFF
--- a/querybook/webapp/ui/Modal/Modal.scss
+++ b/querybook/webapp/ui/Modal/Modal.scss
@@ -58,7 +58,7 @@
             .Modal-content {
                 max-height: inherit;
                 height: auto;
-                overflow-y: hidden;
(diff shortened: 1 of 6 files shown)


querybook/webapp/ui/Modal/Modal.scss
```diff
@@ -58,6 +58,7 @@
             .Modal-content {
                 max-height: inherit;
                 height: auto;
+                overflow-y: hidden;
```
czgu, line 61: this doesn't do anything? since line 62 removes it

meowcodes: 🤦 i meant overflow-x

PR #1072, opened 2022-11-04
feat: Visualize complex Hive column types

baumandm:
This change makes it easier to understand complex Hive types, including `array<>`, `map<>`, `struct<>`, and `uniontype<>`.  It adds a parser that detects these types and converts them into a representative JSON object, which is then visualized using the [react-json-view](https://www.npmjs.com/package/react-json-view) library added in #991.

Currently it only works on these Hive types.  If it's too specific/niche to merge that's fine, we'll maintain it internally on our side.  Just wanted to throw it out there in case it would be useful to anyone else.

Screen Shot 2022-11-04 at 4 25 55 PM Screen Shot 2022-11-04 at 4 19 28 PM

Screen Shot 2022-11-04 at 3 57 14 PM
Screen Shot 2022-11-04 at 4 18 58 PM

### DIFF
--- a/querybook/webapp/lib/utils/complex-types.ts
+++ b/querybook/webapp/lib/utils/complex-types.ts
@@ -1,0 +1,153 @@
+/**
+ * Convert a complex Hive type string to a nested JSON object
+ *
+ * Example: 'struct<date:struct<year:int,month:int,day:int>,hour:int,minute:int,second:int,timeZoneId:string>'
+ * Output: {
+        date: {
+            year: 'int',
+            month: 'int',
+            day: 'int',
+        },
+        hour: 'int',
+        minute: 'int',
+        second: 'int',
+        timeZoneId: 'string',
+    }
+ */
+export function parseType(type: string): any | string {
+    if (type.startsWith('struct<')) {
+        return parseStructType(type);
+    } else if (type.startsWith('array<')) {
+        return parseArrayType(type);
+    } else if (type.startsWith('map<')) {
+        return parseMapType(type);
+    } else if (type.startsWith('uniontype<')) {
+        return parseUnionType(type);
+    } else {
+        return type;
+    }
+}
+
+export function parseStructType(type: string): Record<string, any> | string {
+    const structRegex = /struct<(.*)>/;
+    const structMatch = type.match(structRegex);
+    if (!structMatch) {
+        return type;
+    }
+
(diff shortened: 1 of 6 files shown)


querybook/webapp/lib/utils/complex-types.ts
```diff
@@ -0,0 +1,153 @@
...
+            month: 'int',
+            day: 'int',
+        },
+        hour: 'int',
+        minute: 'int',
+        second: 'int',
+        timeZoneId: 'string',
+    }
+ */
+export function parseType(type: string): any | string {
```
jczhong84, line 17: will there be cases of 
* uppercase, e.g. `STRUCT`
* space between, e.g. `struct <`

and what happens if the parsing fails?

jczhong84 (commented):
* I'm not sure if the JSON format is the best way to help visualize complex hive data types. Json works well for the struct and array type, but not for the map and union types I think. People may get confused what's the actual type is. An alternative way I'm thinking is: how about just simply format the origin type string with indents, e.g.
```
struct<
  date:struct<
    year:int,
    month:int,
    day:int>,
  hour:int,
  minute:int,
  second:int,
  timeZoneId:string
>
```
so that when people are seeing types like `map` or `uniontype`, they will go and check what it is.
* Also I just realized that some really big type string may get truncated. It has a max length of 255 or 256 in our hive metastore. e.g. ```map<string,struct<userid:bigint,matchtypeused:struct<value:int>,matchtypetouseridmap:map<int,bigint>,restrictedusereason:array<struct<value:int>>,isactivated:boolean,experimentname:string,experimentgroup:string,isoptin:boolean,useridtoconfidencescoremap:m```
not sure how you will handle this case?

baumandm:
Thanks for the feedback, I agree JSON isn't the most clear way to visualize some of these types.  

We're using another product that uses a nested column approach for complex types, but it is also confusing when looking at `map<>` and `array<>` types.  Based on the feedback from our users, I think they prefer a confusing visualization over just the raw text.

The advantages were that it was relatively easy to implement and reused the same JSON viewer as the results, providing some consistency for users.  It also is collapsible which is nice for extremely nested types.

I can experiment with different approaches and share some additional options.

Re: the type length limit, I noticed this as well and was going to bring it up later.  The parser degrades as well as it can, but ideally we'd increase the max length.  We too have a significant number of columns synced with truncated types.

PR #1076, opened 2022-11-09
Feat: add enable and list filters for Scheds

alexVasylenko:
This commit adds a bunch of filters for Scheds page.
There are two filters - filter by status and public user lists.
Filter by status has three options: All (default), Enabled and Disabled. A user can choose only one option.
Filter by user lists contains all custom-created lists and a user can select any of them. A user also can choose more than one list.
snap1

### DIFF
--- a/querybook/server/logic/schedule.py
+++ b/querybook/server/logic/schedule.py
@@ -11,6 +11,7 @@ from models.schedule import (
     TaskRunRecord,
 )
 from models.datadoc import DataDoc
+from models.board import BoardItem
 
 DATADOC_SCHEDULE_PREFIX = "run_data_doc_"
 
@@ -193,6 +194,14 @@ def get_scheduled_data_docs_by_user(
 
     if "name" in filters:
         query = query.filter(DataDoc.title.contains(filters.get("name")))
+
+    if filters.get("status") is not None:
+        query = query.filter(TaskSchedule.enabled == filters.get("status"))
+
+    if filters.get("board_ids"):
+        query = query.join(BoardItem, BoardItem.data_doc_id == DataDoc.id).filter(
+            BoardItem.parent_board_id.in_(filters.get("board_ids"))
+        )
 
     count = query.count()
     docs_with_schedules = query.offset(offset).limit(limit).all()
--- a/querybook/webapp/components/DataDocScheduleList/DataDocScheduleList.tsx
+++ b/querybook/webapp/components/DataDocScheduleList/DataDocScheduleList.tsx
@@ -1,17 +1,25 @@
-import React, { useEffect, useMemo, useState } from 'react';
-import { useDispatch, useSelector } from 'react-redux';
-
+import React, {
+    useEffect,
+    useMemo,
+    useState,
+    useRef,
+    useCallback,
+} from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+import { Dispatch, IStoreState } from 'redux/store/types';
 import { getScheduledDocs } from 'redux/scheduledDataDoc/action';
 import { IScheduledDocFilters } from 'redux/scheduledDataDoc/types';
-import { Dispatch, IStoreState } from 'redux/store/types';
 import { Checkbox } from 'ui/Checkbox/Checkbox';
 import { Container } from 'ui/Container/Container';
 import { DebouncedInput } from 'ui/DebouncedInput/DebouncedInput';
 import { Pagination } from 'ui/Pagination/Pagination';
 import { PrettyNumber } from 'ui/PrettyNumber/PrettyNumber';
 import { AccentText, EmptyText } from 'ui/StyledText/StyledText';
+import { IconButton } from 'ui/Button/IconButton';
 
 import { DataDocScheduleItem } from './DataDocScheduleItem';
+import { DataDocSchedsFilters } from './DataDocSchedsFilters';
+import { UpdateFiltersType } from 'const/schedFiltersType';
 
 import './DataDocScheduleList.scss';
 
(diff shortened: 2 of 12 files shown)


querybook/server/logic/schedule.py
```diff
@@ -194,6 +195,14 @@ def get_scheduled_data_docs_by_user(
     if "name" in filters:
         query = query.filter(DataDoc.title.contains(filters.get("name")))
 
+    if "status" in filters:
+        query = query.filter(TaskSchedule.enabled == filters.get("status"))
+
+    if filters.get("list_ids"):
```
czgu, line 201: can we use board_ids internally? (since the internal logic calls it board)

querybook/webapp/components/DataDocScheduleList/DataDocScheduleList.tsx
```diff
@@ -24,34 +45,54 @@ function useDataDocScheduleFiltersAndPagination() {
     } = useSelector((state: IStoreState) => state.scheduledDocs);
 
     const [docName, setDocName] = useState(initFilters.name ?? '');
-    const [scheduledOnly, setScheduledOnly] = useState(
-        initFilters.scheduled_only ?? false
-    );
+
+    const [extraFilters, setExtraFilters] = useState<IScheduledDocFilters>({
+        status: null,
```
czgu, line 50: should these use initFilters as well?

querybook/webapp/components/DataDocScheduleList/DataDocScheduleList.tsx
```diff
@@ -24,34 +45,54 @@ function useDataDocScheduleFiltersAndPagination() {
...
+        list_ids: [],
+        scheduled_only: initFilters.scheduled_only ?? false,
+    });
+
+    const updateFilters = useCallback((params) => {
```
czgu, line 55: missing typing here

querybook/webapp/components/DataDocScheduleList/DataDocScheduleList.tsx
```diff
@@ -66,7 +107,10 @@ function useDataDocWithSchedules(
             getScheduledDocs({
                 paginationPage: page,
                 paginationPageSize: pageSize,
-                paginationFilter: filters,
+                paginationFilter: {
+                    ...filters,
+                    list_ids: filters.list_ids?.map((l) => l.value),
```
czgu, line 112: why not do this processing in filters?

querybook/webapp/components/DataDocScheduleList/DataDocScheduleList.tsx
```diff
@@ -102,6 +146,74 @@ const DataDocScheduleList: React.FC = () => {
...
+    const handleUpdateStatus = React.useCallback(
+        ({ target: { value } }: { target: { value: string } }) => {
+            updateFilters({
+                status: value === '' ? null : value === 'true',
+            });
+        },
+        []
+    );
+
+    const handleUpdateScheduledOnly = React.useCallback((value) => {
```
czgu, line 175: value: boolean?

querybook/webapp/components/DataDocScheduleList/DataDocScheduleList.tsx
```diff
@@ -102,6 +146,74 @@ const DataDocScheduleList: React.FC = () => {
...
+        updateFilters({
+            scheduled_only: value,
+        });
+    }, []);
+
+    const reactSelectStyle = makeReactSelectStyle(true);
```
czgu, line 181: this needs to be outside of the component

querybook/webapp/components/DataDocScheduleList/DataDocScheduleList.tsx
```diff
@@ -102,6 +146,74 @@ const DataDocScheduleList: React.FC = () => {
...
+                    <DataDocScheduleSelectionList label="Status">
+                        <SimpleSelect
+                            value={filters.status}
+                            onChange={handleUpdateStatus}
+                        >
+                            {makeSelectOptions(enabledOptions)}
+                        </SimpleSelect>
+                    </DataDocScheduleSelectionList>
+                    <DataDocScheduleSelectionList label="Lists">
+                        <Select
```
czgu, line 202: separate this Select and the useEffect that loads boards into a BoardSelect component

czgu: also by doing that + adding a memo, the filters can be just ids instead of options

querybook/webapp/components/DataDocScheduleList/DataDocScheduleSelectionList.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import {
+    FormFieldLabelSection,
+    FormFieldInputSection,
+} from 'ui/Form/FormField';
+
+export const DataDocScheduleSelectionList = ({ label, children }) => {
```
czgu, line 7: don't make new components, use FormField instead

querybook/webapp/const/options.tsx
```diff
@@ -0,0 +1,5 @@
+export interface OptionsType {
```
czgu, line 1: can you make the one from EnvironmentsSelect.tsx also use this? and also the one in Select.tsx

querybook/webapp/redux/dataDoc/selector.ts
```diff
@@ -213,3 +213,12 @@ export const queryCellSelector = createSelector(dataDocCellsSelector, (cells) =>
             };
         })
 );
+
+export const queryDataDocFiltersSelector = createSelector(
```
czgu, line 217: this should be under redux/board/selector.ts.

Also you probably don't need this as a selector if you extract the board select into its own component. useMemo would be better because you also need boardById (to map list id to list options)

querybook/webapp/redux/scheduledDataDoc/types.ts
```diff
@@ -8,11 +8,23 @@ import { IDataDoc } from 'const/datadoc';
...
+    status?: boolean;
 }
+
+export interface IScheduledDocFilters extends IBasicScheduledDocFilters {
+    list_ids?: OptionsType[];
+}
+
+export interface ITransformedScheduledDocFilters
+    extends IBasicScheduledDocFilters {
+    list_ids?: string[];
```
czgu, line 25: isn't id numbers?

querybook/webapp/ui/SimpleReactSelect/SimpleReactSelect.tsx
```diff
@@ -36,6 +36,8 @@ export function SimpleReactSelect<T>({
     selectProps = {},
     withDeselect = false,
     clearAfterSelect = false,
+    optionSelector = (val) => val?.value,
+    ...otherParams
```
czgu, line 40: there is no otherParams here from ISimpleReactSelectProps? If you are trying to pass props to the Select component, you need to add it to the interface

czgu (approved):

PR #1082, opened 2022-11-17
feat: add helper task to auto disable unused workflows

czgu:
3 policies:
- disable if user is deactived
- disable if run failed for past N times
- disable if no one has viewed the doc for past N days

Also has notification to owner once its disabled

### DIFF
--- a/querybook/server/tasks/disable_scheduled_docs.py
+++ b/querybook/server/tasks/disable_scheduled_docs.py
@@ -1,36 +1,83 @@
+from datetime import datetime, timedelta
+from enum import Enum
+from typing import Optional
 from app.flask_app import celery
 
 from models.schedule import (
     TaskSchedule,
 )
 from app.db import with_session
+from env import QuerybookSettings
+from const.schedule import TaskRunStatus
+from const.impression import ImpressionItemType
+from lib.notify.utils import notify_user
+from lib.scheduled_datadoc.legacy import convert_if_legacy_datadoc_schedule
 from logic.schedule import (
     DATADOC_SCHEDULE_PREFIX,
     update_task_schedule,
     with_task_logging,
+    get_task_run_records,
+    delete_task_schedule,
 )
 from logic.datadoc import get_data_doc_by_id
 from logic.user import get_user_by_id
+from logic.impression import get_viewers_count_by_item_after_date
 from lib.logger import get_logger
+
 
 logger = get_logger(__name__)
 
 
+class DisablePolicy(Enum):
+    INACTIVE_OWNER = "INACTIVE_OWNER"
+    NO_IMPRESSION_FOR_N_DAYS = "NO_IMPRESSION_FOR_N_DAYS"
+    FAILED_FOR_N_RUNS = "FAILED_FOR_N_RUNS"
+
+
+DISABLE_POLICY_TO_REASON = {
+    DisablePolicy.INACTIVE_OWNER.value: "Disabled due to inactive owner",
+    DisablePolicy.NO_IMPRESSION_FOR_N_DAYS.value: "Disabled due to no impression for {} days",
+    DisablePolicy.FAILED_FOR_N_RUNS.value: "Disabled due to failing for the past {} runs",
+}
+
+
+def disable_policy_to_reason(policy: DisablePolicy, n_runs: int, n_days: int):
+    reason = DISABLE_POLICY_TO_REASON[policy.value]
+    if policy == DisablePolicy.NO_IMPRESSION_FOR_N_DAYS:
+        reason = reason.format(n_days)
+    elif policy == DisablePolicy.FAILED_FOR_N_RUNS:
+        reason = reason.format(n_runs)
+    return reason
+
+
 @with_session
 def get_scheduled_datadoc_tasks(session=None):
-    return (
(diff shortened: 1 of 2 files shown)


querybook/server/tasks/disable_scheduled_docs.py
```diff
@@ -40,27 +87,146 @@ def get_task_owner(task, session):
...
+            should_disable = DisablePolicy.FAILED_FOR_N_RUNS
+
+        if (
+            not should_disable
+            and disable_if_no_impression_for_n_days is not None
+            and check_task_no_impression_for_n_days(
+                task_dict, disable_if_no_impression_for_n_days, session
+            )
+        ):
+            should_disable = DisablePolicy.NO_IMPRESSION_FOR_N_DAYS
```
jczhong84, line 191: might be better to have a separate function to get the disable policy?

querybook/server/tasks/disable_scheduled_docs.py
```diff
@@ -1,36 +1,83 @@
...
+    INACTIVE_OWNER = "INACTIVE_OWNER"
+    NO_IMPRESSION_FOR_N_DAYS = "NO_IMPRESSION_FOR_N_DAYS"
+    FAILED_FOR_N_RUNS = "FAILED_FOR_N_RUNS"
+
+
+DISABLE_POLICY_TO_REASON = {
+    DisablePolicy.INACTIVE_OWNER.value: "Disabled due to deactivated owner",
+    DisablePolicy.NO_IMPRESSION_FOR_N_DAYS.value: "Disabled due to no one viewing the doc for the past {} days",
+    DisablePolicy.FAILED_FOR_N_RUNS.value: "Disabled due to past {} runs all ended up failing",
+}
```
jczhong84, line 41: why not just put the reason as the enum value? so that can use `DisablePolicy.INACTIVE_OWNER.name` for the string of `"INACTIVE_OWNER"` if needed and `DisablePolicy.INACTIVE_OWNER.value` for the reason
```
class DisablePolicy(Enum):
     INACTIVE_OWNER = "Disabled due to deactivated owner""
     NO_IMPRESSION_FOR_N_DAYS = "..."
     FAILED_FOR_N_RUNS = "..."
```

querybook/server/tasks/disable_scheduled_docs.py
```diff
@@ -40,27 +87,146 @@ def get_task_owner(task, session):
...
+                    "task_owner_uid": task_dict["kwargs"].get("user_id"),
+                }
+            )
+
+    if not dry_run:
+        for task_id in tasks_to_delete:
+            delete_task_schedule(task_id, commit=False, session=session)
+
+        for task_to_disable in tasks_to_disable:
+            update_task_schedule(
```
jczhong84, line 212: do we need to save the disable reason back to db for  the schedule?

querybook/server/tasks/disable_scheduled_docs.py
```diff
@@ -40,27 +87,146 @@ def get_task_owner(task, session):
...
+        doc_id = task_to_disable["doc_id"]
+        doc = get_data_doc_by_id(doc_id, session=session)
+
+        # Findout who should be notified of this change
+        # it must be someone who can take action on the doc
+        possible_owners = (
+            [task_to_disable["task_owner_uid"], doc.owner_uid]
+            if task_to_disable["task_owner_uid"]
+            else [doc.owner_uid]
+        ) + [editor.uid for editor in doc.editors if editor.write]
```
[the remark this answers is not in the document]
czgu: because no action could be done for these users?

querybook/server/tasks/disable_scheduled_docs.py
```diff
@@ -40,27 +87,146 @@ def get_task_owner(task, session):
...
+        for possible_owner in possible_owners:
+            user = get_user_by_id(possible_owner, session=session)
+            if not user.deleted:
+                doc_owner = possible_owner
+                break
+
+        if doc_owner is None:
+            continue
+
+        notify_user(
```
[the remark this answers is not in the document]
czgu: hmm i think its fine to skip them for the same reason above

querybook/server/tasks/disable_scheduled_docs.py
czgu: I don't have a good place to put it, but I was thinking to leverage the event logs you built. What do you think?

jczhong84: how about setting the task schedule as disabled instead of deleting it? and put the disable reason in the kwargs?

jczhong84: otherwise, with deleting, after they fixed the data doc, they'll have to recreate the schedule and fill in all the notification, exporters again.

czgu: These task schedules are getting disabled not deleted. The deletion of task schedule only happens to docs that are deleted, in this case no action can be done on them.

querybook/server/tasks/disable_scheduled_docs.py
czgu: good idea 👍

jczhong84 (approved):

PR #1085, opened 2022-11-18
fix: load announcements only if the tab is active

jczhong84:
Found there are a lot of api calls of `/announcement/`, which is about more than 1/3 of total api calls. Here we change it to only call it when the tab is active.


querybook/webapp/components/Announcements/Announcements.tsx
```diff
@@ -24,7 +24,12 @@ export const Announcements: React.FunctionComponent = () => {
         dispatch(querybookUIActions.loadDismissedAnnouncements());
     }, []);
 
-    useInterval(loadAnnouncements, 300000);
+    useInterval(() => {
```
czgu, line 27: `useInterval(useCallback(`
since the function passed is in the deps array of useInterval, it will cause useEffect update upon render

PR #1100, opened 2022-12-08
feat: add datadoc retry with delay on failure

lilyli9:
Adds an option to enable retrying running a Datadoc cell on the event of a failure. Can set the max number of times a cell will retry, and the delay in seconds between each attempt. Retries are performed per-cell, but whenever any cell reaches the max retries limit, the DataDoc is marked failed.

This feature is disabled by default, and all existing schedules continue to work as before.

Screen Shot 2022-12-08 at 11 30 48 AM

Screen Shot 2022-12-08 at 11 07 00 AM

### DIFF
--- a/querybook/server/tasks/run_datadoc.py
+++ b/querybook/server/tasks/run_datadoc.py
@@ -1,4 +1,5 @@
 from celery import chain
+from celery.contrib.abortable import AbortableTask
 
 from app.db import DBSession
 from app.flask_app import celery, socketio
@@ -41,6 +42,7 @@ def run_datadoc_with_config(
     execution_type=QueryExecutionType.SCHEDULED.value,
     # Exporting related settings
     exports=[],
+    retry={"delay_sec": 0, "max_retries": 0, "enabled": False},
     *args,
     **kwargs,
 ):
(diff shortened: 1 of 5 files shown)


querybook/webapp/components/DataDocSchedule/DataDocScheduleForm.tsx
```diff
@@ -435,6 +460,59 @@ const ScheduleNotifactionsForm: React.FC<{
     );
 };
 
+const ScheduleRetryForm: React.FC = () => {
+    const { values } = useFormikContext<IScheduleFormValues>();
+
+    const retryValue = values.kwargs.retry ?? { enabled: false };
+
+    const retryCountOptions = [1, 2, 3, 4, 5].map((retryValue) => ({
```
czgu, line 468: retryCount?

querybook/server/tasks/run_datadoc.py
```diff
@@ -126,7 +131,21 @@ def _start_query_execution_task(
...
+        query_execution_id=query_execution_id,
+        execution_type=execution_type,
+        celery_task=self,
+    )
+    if (
+        retry["enabled"] is True
+        and retry["max_retries"] != 0
+        and query_run_status == QueryExecutionStatus.ERROR.value
+    ):
+        self.retry(countdown=retry["delay_sec"], max_retries=retry["max_retries"])
```
czgu, line 146: this seems like it would infinite retry itself?

czgu (commented):
LGTM! just minor comments

querybook/webapp/components/DataDocSchedule/DataDocScheduleForm.tsx
lilyli9: updated

querybook/server/tasks/run_datadoc.py
lilyli9: Celery should take care of this. Testing shows that on cell run failure, the code retries up to the specified max_retries count. If running the cell continues to fail, and the code has retried the max_retries amount of times, then the code stops trying to run that cell.

PR #1109, opened 2022-12-21
feat: auto add quotes for table/columns

czgu:
If a column is like "foo/bar" then querybook would auto add quotes around the column names, similarly for tables.

For Presto
image

For MySQL
image

also 2 bug fixes:
- If column name has upper case, still match prefix
- make sure getSQLHint always goes to the latest language type

### DIFF
--- a/querybook/webapp/components/QueryEditor/QueryEditor.tsx
+++ b/querybook/webapp/components/QueryEditor/QueryEditor.tsx
@@ -559,13 +559,11 @@ export const QueryEditor: React.FC<
 
         const handleOnFocus = useCallback(
             (editor: CodeMirror.Editor, event) => {
-                autoCompleter.registerHelper();
-
                 if (onFocus) {
                     onFocus(editor, event);
                 }
             },
-            [onFocus, autoCompleter]
+            [onFocus]
         );
 
         const handleOnKeyUp = useCallback(
--- a/querybook/webapp/components/QueryEditor/QueryEditor.tsx
+++ b/querybook/webapp/components/QueryEditor/QueryEditor.tsx
@@ -130,7 +130,7 @@ export const QueryEditor: React.FC<
             language,
             query: value,
         });
-        const autoCompleter = useAutoComplete(
+        const { autoCompleter, autoCompleterRef } = useAutoComplete(
             metastoreId,
             autoCompleteType,
             language,
@@ -559,11 +559,16 @@ export const QueryEditor: React.FC<
 
         const handleOnFocus = useCallback(
             (editor: CodeMirror.Editor, event) => {
+                // This is needed because we could have multiple QueryEditor
+                // instances on the same page
+                // Note that we are using ref here because ReactCodeMirror doesn't
+                // use the new handleOnFocus - it only uses the one on mount
+                autoCompleterRef.current.registerHelper();
                 if (onFocus) {
                     onFocus(editor, event);
                 }
             },
-            [onFocus]
+            [onFocus, autoCompleterRef]
         );
 
         const handleOnKeyUp = useCallback(
--- a/querybook/webapp/components/QueryEditor/QueryEditor.tsx
+++ b/querybook/webapp/components/QueryEditor/QueryEditor.tsx
@@ -130,7 +130,7 @@ export const QueryEditor: React.FC<
             language,
             query: value,
         });
-        const { autoCompleter, autoCompleterRef } = useAutoComplete(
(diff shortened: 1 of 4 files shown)


querybook/webapp/components/QueryEditor/QueryEditor.tsx
```diff
@@ -559,13 +559,11 @@ export const QueryEditor: React.FC<
 
         const handleOnFocus = useCallback(
             (editor: CodeMirror.Editor, event) => {
-                autoCompleter.registerHelper();
```
jczhong84, line 562: why it's not needed?

querybook/webapp/lib/sql-helper/sql-autocompleter.ts
```diff
@@ -233,7 +257,7 @@ export class SqlAutoCompleter {
             (id) => dataSources.dataColumnsById[id].name
         );
         const filteredColumnNames = columnNames.filter((name) =>
-            name.startsWith(prefix)
+            name.toLowerCase().startsWith(prefix)
```
jczhong84, line 260: is `prefix` also lowercased?

querybook/webapp/lib/sql-helper/sql-autocompleter.ts
```diff
@@ -287,10 +313,17 @@ export class SqlAutoCompleter {
                 const tableNames = await this.getTableNamesFromPrefix(prefix);
 
                 for (const tableName of tableNames) {
-                    const match = tableName.match(/^(\w+)\.(\w+)$/);
-                    if (match) {
+                    const schemaTableNames = tableName.split('.');
+
+                    if (schemaTableNames.length === 2) {
```
jczhong84, line 318: I feel like I saw this similar code piece multiple times, might be good to have tiny util function for it? can return an object with parsed schema name and table name, so that

```
const {schemaName, tableName} = parseFullTableName(..)
if (schemaName && tableName) {
   ...
   this.flatQuotationFormatter(tableName)
   ...
}
```

querybook/webapp/lib/sql-helper/sql-autocompleter.ts
```diff
@@ -247,17 +271,19 @@ export class SqlAutoCompleter {
...
             } else if (
                 lineAnalysis.context === 'column' &&
                 lineAnalysis.reference
             ) {
                 results = this.getColumnsFromPrefix(
                     searchStr,
                     lineAnalysis.reference
-                ).map(formatter.bind(null, lineAnalysis.context));
+                )
+                    .map(this.flatQuotationFormatter)
```
jczhong84, line 285: how about just move this `.map(this.flatQuotationFormatter)` into `getColumnsFromPrefix`?

querybook/webapp/components/QueryEditor/QueryEditor.tsx
czgu: actually it is needed, otherwise query hint wouldn't work. I changed back the code and added more comments

querybook/webapp/lib/sql-helper/sql-autocompleter.ts
czgu: yes, in this line
```
searchStr = token.string.toLowerCase();
```

querybook/webapp/lib/sql-helper/sql-autocompleter.ts
czgu: hmm it is just `.split('.')` and used in 2 places, i think we can make a function for it if it is more common in future

querybook/webapp/lib/sql-helper/sql-autocompleter.ts
czgu: i moved all of the quotation formatting logic to the formatter level so it would not interfere with display

querybook/webapp/components/QueryEditor/QueryEditor.tsx
```diff
@@ -130,7 +130,7 @@ export const QueryEditor: React.FC<
             language,
             query: value,
         });
-        const autoCompleter = useAutoComplete(
+        const { autoCompleter, autoCompleterRef } = useAutoComplete(
```
jczhong84, line 133: seems `autoCompleter` is never used now, why it needs to be returned from `useAutoComplete`? and it's actually equals to autoCompleterRef.current

jczhong84 (approved):

PR #1118, opened 2023-01-06
Add some frontend view/click events

jczhong84:
This is the first batch of events we're interested for tracking. 

Also fixed a bug of creating datadoc from adhoc with templated variables.


czgu (dismissed):

querybook/webapp/components/Search/SearchOverview.tsx
```diff
@@ -112,6 +112,18 @@ export const SearchOverview: React.FC<ISearchOverviewProps> = ({
     const results = resultByPage[currentPage] || [];
     const isLoading = !!searchRequest;
 
+    // Log search results
+    useEffect(() => {
+        if (!isLoading && !!searchString.length && !!results.length) {
```
czgu, line 117: can you use searchString.length > 0 and results.length >0, seems easier to understand than `!!`

czgu (approved):

PR #1119, opened 2023-01-06
feat: validation now works for templated query

czgu:
- fixed an issue where templated_variables kept getting regenerated
- make templated queries also get validated
- make templated query render modal also show validation errors with highlights
image


querybook/webapp/components/TemplateQueryView/TemplatedQueryView.tsx
```diff
@@ -41,6 +104,13 @@ export const TemplatedQueryView: React.FC<ITemplatedQueryViewProps> = ({
         )
     );
 
+    const { validationErrorHighlights, queryValidationErrors, isValidating } =
+        useValidateQuery(
+            renderedQuery,
```
jczhong84, line 109: Instead of 2 api calls here, I'm wondering if it'll be better to have the `renderTemplatedQuery` api to accept another parameter to indicate if it should also do the validation on the rendered query and return the validation result along with the rendered query?

czgu: hm I don't want to tie the two API calls together too much. One is for rendering the template and the other is to do validation check (only if the validator is configured). Also in future, I am thinking that we can make template rendering different if it is for validation so that we can map back the validation result from rendered query to unrendered query

jczhong84 (approved):

PR #1122, opened 2023-01-09
fix: Add disabled indicators to Schedules list

baumandm:
This PR provides some visual cues for disabled schedules in the Schedules list: 

- Shows `Next Run: Disabled` instead of the would-be timestamp (if it was enabled)
- The DataDoc name heading is lightened

Screenshot 2023-01-09 at 16-34-18 Querybook


czgu (approved):
awesome!

PR #1127, opened 2023-01-13
chore: update richtext editor to hooks

czgu:
- RichTextEditor and RichTextEditorToolbar are now in hooks
- No other logic change except componentDidCatch was removed

### DIFF
--- a/querybook/webapp/components/SearchAndReplace/DraftJsSearchHighlighter.tsx
+++ b/querybook/webapp/components/SearchAndReplace/DraftJsSearchHighlighter.tsx
@@ -5,13 +5,13 @@ import scrollIntoView from 'smooth-scroll-into-view-if-needed';
 import { makeSearchHighlightDecorator } from 'components/SearchAndReplace/SearchHighlightDecorator';
 import { ISearchAndReplaceContextType } from 'context/searchAndReplace';
 import { LinkDecorator } from 'lib/richtext';
-import { RichTextEditor } from 'ui/RichTextEditor/RichTextEditor';
+import type { IRichTextEditorHandles } from 'ui/RichTextEditor/RichTextEditor';
 
 export const DraftJsSearchHighlighter: React.FC<{
-    editor: RichTextEditor;
+    editorRef: React.MutableRefObject<IRichTextEditorHandles>;
     cellId: number;
     searchContext: ISearchAndReplaceContextType;
-}> = ({ editor, cellId, searchContext }) => {
+}> = ({ editorRef, cellId, searchContext }) => {
     const {
         searchState: {
             searchResults,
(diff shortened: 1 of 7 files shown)


querybook/webapp/ui/RichTextEditor/RichTextEditor.tsx
```diff
@@ -45,523 +55,562 @@ export interface IRichTextEditorState {
...
+            decorator,
+
+            onChange,
+            onKeyDown,
+            onFocus,
+            onBlur,
+        },
+        ref
+    ) => {
+        const [editorState, setEditorState, editorStateRef] = useStateWithRef(
```
jczhong84, line 130: dont quite get why we need `useStateWithRef` here.
1. I dont see `editorState` is used in any `useEffect`'s deps array.
2. If we'd like to not trigger useEffect update when editorState changes, why not just use `useRef` instead of state?

querybook/webapp/ui/RichTextEditor/RichTextEditor.tsx
```diff
@@ -45,523 +55,562 @@ export interface IRichTextEditorState {
...
-    public componentDidMount() {
-        if (this.props.autoFocus) {
-            this.focus();
-        }
-    }
+                // stop event progation if the event is
+                // either handled by default behavior or custom behaivor
+                if (handled) {
+                    e.stopPropagation();
+                    e.preventDefault();
```
jczhong84, line 518: use `stopPropagationAndDefault`?

querybook/webapp/components/SearchAndReplace/DraftJsSearchHighlighter.tsx
```diff
@@ -23,23 +23,26 @@ export const DraftJsSearchHighlighter: React.FC<{
...
-                decorator: new DraftJs.CompositeDecorator(decorators),
-            });
+            editorRef.current.setEditorState(
+                DraftJs.EditorState.set(editorRef.current.getEditorState(), {
+                    decorator: new DraftJs.CompositeDecorator(decorators),
+                })
+            );
         }
-    }, [shouldHighlight, editor, searchString, searchOptions]);
+    }, [shouldHighlight, editorRef, searchString, searchOptions]);
```
jczhong84, line 45: ref will not change, it does nothing in the deps array I guess?

czgu: its just to suppress the hooks warning, since it doesn't change it shouldn't affect anything

querybook/webapp/ui/RichTextEditor/RichTextEditor.tsx
czgu: its to suppress frequent updates in useImperativeHandles and useCallback

jczhong84 (approved):

PR #1134, opened 2023-01-24
feat: add metastore loader config

jczhong84:
As we may sync more metadata types from metastore, like owners, tags, here we are adding a config for the metastore loader so that we can manage whether the metadata should be read-only or written back to metastore.

By default, it will keep the current behavior, which is only writing into querybook db.

### DIFF
--- a/querybook/server/const/metastore.py
+++ b/querybook/server/const/metastore.py
@@ -4,3 +4,29 @@ from enum import Enum
 class DataTableWarningSeverity(Enum):
     WARNING = 0
     ERROR = 1
+
+
+class MetadataType(Enum):
+    TABLE_DESCRIPTION = "table_description"
+    COLUMN_DESCRIPTION = "column_description"
+    OWNER = "owner"
+    TAG = "tag"
+    DOMAIN = "domain"
+
+
+class MetadataMode(Enum):
+    READ_ONLY = "read_only"  # metadata will be read-only on Querybook UI
+    WRITE_BACK = "write_back"  # metadata will be written back to metastore on saving
+
+
+class MetastoreLoaderConfig:
+    """Config to control if the metadata can be edited or written back to metastore.
+    The default behavior of a metadata type (not specifed in the config) is that
+    it will be editable and only saved in qurybook db.
+    """
+
+    def __init__(self, config: dict[MetadataType, MetadataMode]):
+        self._config = config
+
+    def to_dict(self):
+        return {key.value: value.value for (key, value) in self._config.items()}
--- a/querybook/server/const/metastore.py
+++ b/querybook/server/const/metastore.py
@@ -11,25 +11,31 @@ class MetadataType(Enum):
     COLUMN_DESCRIPTION = "column_description"
     OWNER = "owner"
     TAG = "tag"
-    DOMAIN = "domain"
 
 
 class MetadataMode(Enum):
     # Metadata will be read-only on Querybook UI and it will redirect to the metastore link for editing.
     READ_ONLY = "read_only"
 
-    # On saving, metadata will only be written back querybook db. This is the default mode if not specified.
-    WRITE_BACK = "write_back"
+    # On saving, metadata will only be written to querybook db. This is the default mode if not specified.
+    WRITE_LOCAL = "write_local"
 
     # On saving, metadata will be written back to metastore, as well as querybook db
-    WRITE_THROUGH = "write_through"
+    WRITE_BACK = "write_back"
 
 
 class MetastoreLoaderConfig:
     """Config to set the read/write mode (MetadataMode) for each metadata."""
 
+    _default_config = {
+        MetadataType.TABLE_DESCRIPTION: MetadataMode.WRITE_LOCAL,
+        MetadataType.COLUMN_DESCRIPTION: MetadataMode.WRITE_LOCAL,
+        MetadataType.OWNER: MetadataMode.WRITE_LOCAL,
+        MetadataType.TAG: MetadataMode.WRITE_LOCAL,
+    }
+
     def __init__(self, config: dict[MetadataType, MetadataMode]):
-        self._config = config
+        self._config = {**self._default_config, **config}
 
     def to_dict(self):
         return {key.value: value.value for (key, value) in self._config.items()}
--- a/querybook/server/const/metastore.py
+++ b/querybook/server/const/metastore.py
@@ -15,15 +15,18 @@ class MetadataType(Enum):
 
 
 class MetadataMode(Enum):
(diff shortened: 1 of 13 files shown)


querybook/webapp/components/DataTableView/DataTableView.tsx
```diff
@@ -158,6 +181,7 @@ class DataTableViewComponent extends React.PureComponent<
     @bind
     public makeOverviewDOM() {
         const { table, tableName, tableColumns, tableWarnings } = this.props;
+        console.log(this.props);
```
czgu, line 184: remove console

querybook/webapp/const/metastore.ts
```diff
@@ -1,8 +1,23 @@
...
+
+// Keep it in sync with MetadataMode in server/const/metastore.py
+export enum MetadataMode {
+    READ_ONLY = 'read_only',
+    WRITE_BACK = 'write_back',
+}
 export interface IQueryMetastore {
     id: number;
     name: string;
+    config: { [key in MetadataType]: MetadataMode };
```
czgu, line 20: use Record<MetadataType, MetadataMode>? the key doesn't have much significance

querybook/webapp/ui/EditableTextField/EditableTextField.tsx
```diff
@@ -17,11 +17,12 @@ export interface IEditableTextFieldProps {
     className?: string;
     readonly?: boolean;
     placeholder?: string;
+    onEdit?: () => void;
```
czgu, line 20: i think its better if we just have a readonly richtext field with a redirect button. Most of the code in EditableTextField is useless if onEdit is provided

querybook/webapp/components/DataTableViewColumn/DataTableColumnCard.tsx
```diff
@@ -17,6 +17,7 @@ import './DataTableColumnCard.scss';
 
 interface IProps {
     column: IDataColumn;
+    onEdit: () => void;
```
czgu, line 20: im seeing onEditMetadata and onEdit, can you make the naming consistent?

querybook/webapp/components/DataTableView/DataTableView.tsx
```diff
@@ -168,6 +192,9 @@ class DataTableViewComponent extends React.PureComponent<
                 onTabSelected={this.onTabSelected}
                 updateDataTableDescription={this.updateDataTableDescription}
                 onExampleFilter={this.handleExampleFilter}
+                onEditMetadata={this.getOnEditMetadata(
```
czgu, line 195: onBeforeEditDescription or onEditDescriptionRedirect?

querybook/webapp/components/DataTableViewColumn/DataTableViewColumn.tsx
```diff
@@ -15,6 +15,7 @@ export interface IDataTableViewColumnProps {
         columnId: number,
         description: ContentState
     ) => any;
+    onEditMetadata: () => void;
```
czgu, line 18: typing should be

```
onEditMetadata?: Nullable<() => Promise<any>>

File: querybook/server/const/metastore.py [hunk 7]
```diff
@@ -4,3 +4,29 @@
...
+class MetadataType(Enum):
+    TABLE_DESCRIPTION = "table_description"
+    COLUMN_DESCRIPTION = "column_description"
+    OWNER = "owner"
+    TAG = "tag"
+    DOMAIN = "domain"
+
+
+class MetadataMode(Enum):
+    READ_ONLY = "read_only"  # metadata will be read-only on Querybook UI
```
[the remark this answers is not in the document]
jczhong84: My thought was: if a metadata type key is not in the loader config, then it will be editable and only saved in querybook db, which is today's behavior.

I can also add another option for it, and make it default
read_only: 
write_back:  default
write_through: write to querybook db and also write back to metastore

querybook/server/datasources/metastore.py
```diff
@@ -102,6 +102,23 @@ def get_table_by_name(
...
+def get_table_metastore_link(
+    schema_name,
+    table_name,
+    metastore_id,
+    metadata_type,
+):
+    verify_metastore_permission(metastore_id)
+    metastore_loader = get_metastore_loader(metastore_id)
+
+    return metastore_loader.get_metastore_link(
```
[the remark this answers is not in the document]
jczhong84: good point!

querybook/server/models/admin.py
```diff
@@ -157,7 +157,14 @@ class QueryMetastore(CRUDMixin, Base):
     acl_control = sql.Column(sql.JSON, default={}, nullable=False)
 
     def to_dict(self):
-        return {"id": self.id, "name": self.name}
+        from lib.metastore import get_metastore_loader_class_by_name
```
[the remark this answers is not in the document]
jczhong84: yep

querybook/webapp/ui/EditableTextField/EditableTextField.tsx
jczhong84: I think it's similar as when readonly is true, most of the code is also useless. are you saying we should render a readonly rich text for both of them?

querybook/server/datasources/metastore.py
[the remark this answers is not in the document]
jczhong84: I thought it over, perhaps we dont need to restrict the metastore link only for read-only mode . The metastore link exists no matter it's read-only or not, but as long as there is a link. For now, we only use it for redirection. but maybe we could still put the link some where even when it will write back to the metastore.

querybook/webapp/components/DataTableView/DataTableView.tsx
jczhong84: I think it's not "before" actually, but it's "on editing" when clicking, although it will go to another site for editing. 
i'll rename is as `onEditTableDescription`

querybook/server/const/metastore.py
```diff
@@ -4,3 +4,32 @@
...
     WARNING = 0
     ERROR = 1
+
+
+class MetadataType(Enum):
+    TABLE_DESCRIPTION = "table_description"
+    COLUMN_DESCRIPTION = "column_description"
+    OWNER = "owner"
+    TAG = "tag"
+    DOMAIN = "domain"
```
czgu, line 14: lets remove DOMAIN in this PR for now, since this metadata entity does not actually exist.
Once we do treat domain as something in querybook, we can add it back in

czgu (approved):

PR #1138, opened 2023-01-31
fix: stay on current env after deleting doc

lilyli9:
This fixes an issue where deleting a DataDoc in a non-default environment causes the UI to jump to the default environment instead of remaining in the current environment. The issue seems to be related to QB reloading after attempting to navigate to '{currentEnv.name}/datadoc' after deleting a datadoc. The change included in this PR avoids the reload while also refreshing the page to be able to see that the datadoc was deleted.


czgu (approved):
the issue is we are redirecting to root when deleting doc instead redirecting to env:
```
<Route path="/:env/datadoc/" exact={true}>
            <Redirect to="/" />
        </Route>
```
just leaving it as a note here. No change needed.

Thanks for fixing this bug!

PR #1140, opened 2023-02-01
chore: Move QueryViewNavigator to hooks

czgu:
no logic change

### DIFF
--- a/querybook/webapp/components/QueryViewNavigator/QueryViewNavigator.tsx
+++ b/querybook/webapp/components/QueryViewNavigator/QueryViewNavigator.tsx
@@ -1,8 +1,9 @@
-import { bind, debounce } from 'lodash-decorators';
-import React from 'react';
-import { connect } from 'react-redux';
+import { debounce } from 'lodash';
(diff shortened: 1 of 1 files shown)


querybook/webapp/components/QueryViewNavigator/QueryViewNavigator.tsx
```diff
@@ -19,148 +20,138 @@ import { QueryViewFilter } from './QueryViewFilter';
...
+
+    useEffect(() => {
+        initializeFromQueryParam();
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+
+    useEffect(() => {
+        setupPolling(queryResults);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [queryResults]);
```
jczhong84, line 109: why does it not need to depend on setupPolling?

czgu: We only need to poll when the queries change, when setupPoll change we don't need to do the polling again

jczhong84 (approved):

PR #1147, opened 2023-02-04
update querybook helm charts

okayhooni:
- add ingress template to make http/https endpoint on webserver component
- add extraEnv props. to inject common environment variables to all the pods of components
- fix pod assigning configs on `values.yaml` to valid position (all the pod templates actually has same global configs)

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -15,3 +15,5 @@ maintainers:
     email: yongchana@gmail.com
   - name: Jaehyeok Lee
     email: q32145@gmail.com
+  - name: Changhoon Oh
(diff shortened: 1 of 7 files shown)


czgu (commented):
LGTM, is there any particular reason you are changing port to 8081 instead of 10001? would this break anyone using the previous version?

okayhooni:
> LGTM, is there any particular reason you are changing port to 8081 instead of 10001? would this break anyone using the previous version?

this container port change was related to our specific k8s setup. but it works well without any side-effect.

I made this target container port to templated value with default = 10001

czgu (approved):

PR #1151, opened 2023-02-08
feat: add event logging for query execution progress

jczhong84:
During the query execution, we'll log the execution progress every min with elapsed time and progress percentage.


jczhong84:
abandoned

PR #1159, opened 2023-02-11
feat: Add ability to cancel dead queries

czgu:
previously, we would just send abort signal to the execution. But if the query is not running in celery, then nothing would be done. This PR adds more checks on the API side to update query status if the execution itself is gone

### DIFF
--- a/querybook/server/datasources/query_execution.py
+++ b/querybook/server/datasources/query_execution.py
@@ -176,13 +176,19 @@ def cancel_query_execution(query_execution_id):
 
     task = run_query_task.AsyncResult(execution.task_id)
 
-    if task.state == "PENDING":
-        # Task is unknown or haven't delivered to worker yet
+    if task.state in (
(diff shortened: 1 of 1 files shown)


querybook/server/datasources/query_execution.py
```diff
@@ -116,20 +123,65 @@ def get_query_execution(query_execution_id):
...
-            task = run_query_task.AsyncResult(execution_dict["task_id"])
-            if task is not None:
-                task.abort()
+    if not execution.task_id:
+        cancel_query_and_notify()
+        return
+
+    task = run_query_task.AsyncResult(execution.task_id)
+
+    if task.state == "PENDING":
```
jczhong84, line 176: celery task has another `RECEIVED` state, which also can be revoked. would be nice to add this case as well.

jczhong84 (approved):

jczhong84 (approved):

PR #1163, opened 2023-02-14
fix: filter null store values

lilyli9:
This fixes an issue where having a query engine with a missing or deleted metastore causes the Table Upload UI to crash on the "Table Spec" section.


czgu (approved):
thanks!

PR #1165, opened 2023-02-15
feat: add table custom properties

jczhong84:
Tables from metatore may have some custom properties, here we're adding this column to the `data_table_information` table and load it from metastore.


querybook/server/lib/metastore/base_metastore_loader.py
```diff
@@ -65,6 +65,9 @@ class DataTable(NamedTuple):
     # Arrays of partition keys
     partition_keys: List[str] = []
 
+    # Custom properties
+    custom_properties: dict = None
```
czgu, line 69: `dict[str, str]`?

czgu:
can you bump the minor version since there is a db change

czgu (approved):

PR #1169, opened 2023-02-16
feat: add column level tag support

jczhong84:
This PR will add column level tag support. For now, it will only show the tags synced from metastore, but not allow users to add/edit a tag through UI

Snip20230215_11


querybook/server/lib/metastore/base_metastore_loader.py
```diff
@@ -329,24 +329,34 @@ def _create_table_table(
             )
 
             for column in columns:
-                create_column(
+                column_id = create_column(
```
czgu, line 332: just to double check, i thought flushing the row would generate the id?

czgu: https://stackoverflow.com/questions/28242523/get-inserted-key-before-commit-session

querybook/server/logic/tag.py
```diff
@@ -16,6 +16,17 @@ def get_tag_by_table_id(table_id, session=None):
     )
 
 
+@with_session
+def get_tag_by_column_id(column_id: int, session=None):
```
czgu, line 20: get_tags

querybook/server/logic/tag.py
```diff
@@ -128,3 +139,41 @@ def create_table_tags(
         session.commit()
     else:
         session.flush()
+
+
+@with_session
+def create_column_tags(
+    column_id: int = None,
```
czgu, line 146: column_id should be required

querybook/webapp/components/DataTableViewColumn/DataTableColumnCard.tsx
```diff
@@ -69,6 +86,11 @@ export const DataTableColumnCard: React.FunctionComponent<IProps> = ({
                                 />
                             </KeyContentDisplay>
                         )}
+                        {columnTags && columnTags.length > 0 && (
+                            <KeyContentDisplay keyString="Tags">
```
czgu, line 90: better if you do
```
    const tagsDOM = (columnTags || []).map((tag) => (
                <TableTag tag={tag} readonly={true} key={tag.id} mini={true} />
            ));

   tagsDOM && (
      <KeyContentDisplay keyString="Tags"><div className="DataTableTags flex-row">{tagsDOM}</div>
      </KeyContentDisplay>

  )

```

jczhong84: tagsDom is not falsey when columnTags is []

querybook/server/lib/metastore/base_metastore_loader.py
jczhong84: yep, flush actually works

querybook/webapp/components/DataTableViewColumn/DataTableColumnCard.tsx
czgu: oh then tagsDOm.length

querybook/server/datasources/tag.py
```diff
@@ -11,10 +12,19 @@
...
+    verify_data_table_permission(table_id)
+    return logic.get_tags_by_table_id(table_id=table_id)
+
+
+@register(
+    "/column/<int:column_id>/tag/",
+    methods=["GET"],
+)
+def get_tags_by_column_id(column_id: int):
+    column = metastore_logic.get_column_by_id(column_id)
```
czgu, line 25: can you add `verify_data_column_permission` so the checks are in 1 place?

czgu (approved):

PR #1170, opened 2023-02-17
feat: load data table ownership from metastore

jczhong84:
Add support of loading table ownerships from metastore.  For the table ownership, we also bring in a concept of `ownership type`, you can provide types like table creator, business owner, technical owner and etc.  Ownership type is only supported when syncing from metastore for now. If metatore is not configured to load owners, the behavior will remain the same as today.

For each type, it will be rendered in a separate line, like
Snip20230217_15

If you only like to sync owners but without providing types, by default it will look like today
Snip20230217_16


querybook/server/logic/metastore.py
```diff
@@ -393,6 +398,33 @@ def create_table_ownership(table_id, uid, commit=True, session=None):
...
+    table_id: int, owners: list[DataOwner] = [], commit=True, session=None
+):
+    """This function is used for loading owners from metastore."""
+    # delete all the ownerships of the table first
+    session.query(DataTableOwnership).filter_by(data_table_id=table_id).delete()
+
+    for owner in owners:
+        user = get_user_by_name(owner.username)
+        if not user:
+            LOG.error(
```
czgu, line 412: gotta think of internal cases where the name is invalid

querybook/webapp/components/DataTableView/DataTableHeader.tsx
```diff
@@ -57,11 +57,13 @@ export const DataTableHeader: React.FunctionComponent<IDataTableHeader> = ({
         [table.id, userInfo.uid]
     );
 
-    const { username, tableOwnerships } = useSelector((state: IStoreState) => ({
-        username: state.user.userInfoById[userInfo.uid].username,
-        tableOwnerships:
-            state.dataSources.dataTableOwnershipByTableId[table.id],
-    }));
+    const { username, tableOwnerships = [] } = useSelector(
```
czgu, line 60: useShallowSelector

querybook/webapp/components/DataTableView/DataTableHeader.tsx
```diff
@@ -198,7 +236,9 @@ export const DataTableHeader: React.FunctionComponent<IDataTableHeader> = ({
         <div className="DataTableHeader flex-column mb16">
             {topDOM}
             {titleDOM}
-            {ownerDOM}
+            {metastore.config[MetadataType.OWNER] !== MetadataMode.WRITE_LOCAL
```
czgu, line 239: how does add ownership works now with ownership type?

querybook/server/logic/metastore.py
jczhong84: If the name is invalid, then it can't be found, this `if not uesr` will catch it. or are you talking about even the name is invalid, we should still show it?

querybook/webapp/components/DataTableView/DataTableHeader.tsx
jczhong84: ownership type only be supported when syncing from metastore for now. If metatore is not configured to load owners, the behavior will remain the same as today.

querybook/server/lib/metastore/base_metastore_loader.py
```diff
@@ -57,6 +58,35 @@ def get_metastore_link(
...
+                "name": "BUSINESS_OWNER",
+                "display_name": "Owner",
+                "description": "Person or group who is responsible for business related aspects of the table",
+            },
+        ]
+        """
+        return [
+            {
+                "name": None,
+                "display_name": "Owners",
```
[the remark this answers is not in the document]
jczhong84: This is the default behavior like today, which is `Owners`. or should we change it to `Owner` as well?

czgu (approved):

czgu (approved):

PR #1176, opened 2023-02-23
feat: load user groups 

jczhong84:
* add the suppport of creating and updating user groups
* add latest_partitions and earliest_partitions as metastore may provide it directly instead of all partitions.


baumandm:
Question, why is this tied to the metastore loader and not a new plugin type?  We use Hive, and while we can implement the `get_all_user_groups` method and have it connect to Ldap, it doesn't seem like it really fits there.  We also have multiple Hive metastores, but we would only want to schedule the user sync for one of them.

If you have a different metastore that provides user/groups information directly, I can see how this would make more sense.

jczhong84:
makes sense, I'll remove the sync job and decouple it from metastore loader, but only keep the help functions in the PR

querybook/server/const/metastore.py
```diff
@@ -81,6 +83,15 @@ class DataColumn(NamedTuple):
     tags: list[DataTag] = []
 
 
+class UserGroup(NamedTuple):
```
czgu, line 86: move this to a separate const field

querybook/server/const/metastore.py
```diff
@@ -56,6 +56,8 @@ class DataTable(NamedTuple):
 
     # Json arrays of partitions
     partitions: list = []
+    earliest_partitions: list = None
```
czgu, line 59: these are list[str] right? i forgot why partitions' typing is only list lol

querybook/server/logic/user.py
```diff
@@ -144,6 +140,52 @@ def delete_user(uid, session=None):
...
+        # update the group
+        group = User.update(
+            id=group.id,
+            fields=fields,
+            commit=commit,
+            session=session,
+        )
+
+    # get new member user ids by name
+    new_group_members = (
```
czgu, line 171: i got a bit confused on "new" group members. It looks like they could also be not exist and already exist?

Have you considered instead of deleting all, only delete/create the deltas? it is more complicated but we would be able to use the created_at field to retain when the user joined the group

querybook/server/logic/user.py
```diff
@@ -144,6 +140,65 @@ def delete_user(uid, session=None):
...
+    existing_group_member_ids = [m.uid for m in existing_group_members]
+
+    # get the latest member user ids by name
+    group_members = (
+        session.query(User).filter(User.username.in_(user_group.members)).all()
+    )
+    group_member_ids = [m.id for m in group_members]
+
+    # delete group members not in the group anymore
+    session.query(UserGroupMember).filter(UserGroupMember.gid == group.id).filter(
```
czgu, line 183:
```py
existing_group_member_ids = set([m.uid for m in existing_group_members])
group_member_ids = set([m.id for m in group_members])

members_to_delete = list(existing_group_member_ids - group_member_ids)
members_to_add =  list(group_member_ids - existing_group_member_ids)

if members_to_delete:
    ...
if members_to_add:
   ...
```

czgu (approved):

PR #1179, opened 2023-02-24
feat: update delete cell confirmation popup

lilyli9:
This adds more detail to the delete datadoc cell confirmation page to help users check which cell will be deleted. The confirmation page new details include the type of cell (text/query/chart), the name of the cell if it exists, and a query cell preview for query datadoc cells. This cell preview is uneditable, and shows a scroll bar for longer queries. 

Screen Shot 2023-02-24 at 11 46 36 AM
Screen Shot 2023-02-24 at 11 46 24 AM
Screen Shot 2023-02-24 at 11 46 47 AM

***Currently the cell preview has a known issue where queries less than 10 lines will overlap with the query line count, depicted in the below screenshot. We were unable to fix this overlap, but otherwise the cell preview works as expected.***

Screen Shot 2023-02-24 at 11 47 23 AM

### DIFF
--- a/querybook/webapp/components/DataDoc/DataDoc.tsx
+++ b/querybook/webapp/components/DataDoc/DataDoc.tsx
@@ -60,6 +60,7 @@ import { Message } from 'ui/Message/Message';
 
 import { DataDocCellControl } from './DataDocCellControl';
 import { DataDocContentContainer } from './DataDocContentContainer';
+import { DataDocDeletePreview } from './DataDocDeletePreview';
 import { DataDocError } from './DataDocError';
 import { DataDocHeader } from './DataDocHeader';
 import { DataDocLoading } from './DataDocLoading';
@@ -427,9 +428,18 @@ class DataDocComponent extends React.PureComponent<IProps, IState> {
                     }
                 };
                 if (shouldConfirm) {
+                    let cellName = '';
+                    if ('title' in cell.meta) {
+                        cellName = cell.meta.title
+                            ? ` "${cell.meta.title}"`
+                            : cellName;
+                    }
                     sendConfirm({
-                        header: 'Delete Cell?',
-                        message: 'Deleted cells cannot be recovered',
+                        header: `Delete ${
+                            cell.cell_type[0].toUpperCase() +
+                            cell.cell_type.slice(1).toLowerCase()
+                        } Cell${cellName}?`,
+                        message: <DataDocDeletePreview cell={cell} />,
                         onConfirm: deleteCell,
                         onHide: resolve,
                         confirmColor: 'cancel',
--- a/querybook/webapp/components/DataDoc/DataDocDeletePreview.tsx
+++ b/querybook/webapp/components/DataDoc/DataDocDeletePreview.tsx
@@ -1,0 +1,31 @@
+import React, { useEffect, useState } from 'react';
+
+import { IDataCell } from 'const/datadoc';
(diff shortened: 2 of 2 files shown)


querybook/webapp/components/DataDoc/DataDoc.tsx
```diff
@@ -427,9 +428,18 @@ class DataDocComponent extends React.PureComponent<IProps, IState> {
                     }
                 };
                 if (shouldConfirm) {
+                    let cellName = '';
```
czgu, line 431: this part is same as 
```ts
cellName = ` ${cell.meta.title ?? ''}`
```

querybook/webapp/components/DataDoc/DataDoc.tsx
```diff
@@ -427,9 +428,18 @@ class DataDocComponent extends React.PureComponent<IProps, IState> {
...
+                    if ('title' in cell.meta) {
+                        cellName = cell.meta.title
+                            ? ` "${cell.meta.title}"`
+                            : cellName;
+                    }
                     sendConfirm({
-                        header: 'Delete Cell?',
-                        message: 'Deleted cells cannot be recovered',
+                        header: `Delete ${
```
czgu, line 438: `Delete ${capitalize(cell.cell_type)}`

querybook/webapp/components/DataDoc/DataDocDeletePreview.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { ThemedCodeHighlightWithMark } from 'ui/CodeHighlight/ThemedCodeHighlightWithMark';
+
+interface IProps {
+    cell: IDataCell;
+}
+
+export const DataDocDeletePreview: React.FunctionComponent<IProps> = ({
+    cell,
+}) => {
+    const [query, setQuery] = useState<string>();
```
czgu, line 13: why do you need query as a state? vs
```
const query = cell.context as string
```

querybook/webapp/components/DataDoc/DataDocDeletePreview.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useEffect, useState } from 'react';
+
+import { IDataCell } from 'const/datadoc';
+import { ThemedCodeHighlightWithMark } from 'ui/CodeHighlight/ThemedCodeHighlightWithMark';
+
+interface IProps {
+    cell: IDataCell;
+}
+
+export const DataDocDeletePreview: React.FunctionComponent<IProps> = ({
```
czgu, line 10: probably don't need another component for this

lilyli9: Responding to this comment and the previous comment, the separate component was added because without it CodeMirror wouldn't show the query contents when the delete confirmation first loaded. The fix we found was to force it to rerender by changing the state variable with useEffect, which isn't possible without a separate component.

querybook/webapp/components/DataDoc/DataDoc.tsx
lilyli9: Structured this part in this way since I was getting the below linting error. 
```
Property 'title' does not exist on type 'IDataQueryCellMeta | IDataCellMetaBase | IDataChartCellMeta'.
  Property 'title' does not exist on type 'IDataCellMetaBase'.
```

PR #1184, opened 2023-02-28
feat: find the closed color in ColorPalette

jczhong84:
For unknown hex tag colors, we'll try to find the closest color in the ColorPalette.

Also fixed an issue of the owner type, which needs to return a dict instead of a named tuple.


querybook/webapp/lib/utils/color.ts
```diff
@@ -0,0 +1,34 @@
+import { IColorPalette } from 'const/chartColors';
+
+const convertHexToRGB = (hexColor) => {
```
czgu, line 3: need typing here

querybook/server/models/admin.py
```diff
@@ -164,7 +164,7 @@ def to_dict(self):
             "id": self.id,
             "name": self.name,
             "config": loader_class.loader_config.to_dict(),
-            "owner_types": loader_class.get_table_owner_types(),
+            "owner_types": [t._asdict() for t in loader_class.get_table_owner_types()],
```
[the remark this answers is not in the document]
jczhong84: this `owner_types` is nested in the metastore model's `to_dict()`, but not a direct api response. json.py will only handle the top level objects.

querybook/server/logic/tag.py
```diff
@@ -119,7 +120,7 @@ def create_table_tags(
         meta = {
             "type": tag.type,
             "tooltip": tag.description,
-            "color": tag.color,
+            "color": closest_color(tag.color),
```
czgu, line 123: if we are assuming tag.color has format `#AABBCC` then can you add a comment in DataTag?

querybook/webapp/ui/Tag/Tag.tsx
```diff
@@ -55,7 +55,9 @@ const StyledColorTag = styled.span.attrs<{
         let bgColor = 'var(--bg-light)';
         let textColor = 'var(--text-light)';
         if (props.color) {
-            const color = ColorPalette.find((c) => c.name === props.color);
+            const color = ColorPalette.find(
+                (c) => c.name === props.color || c.color === props.color
```
czgu, line 59: `c.color === props.color` is not needed

czgu (approved):

PR #1188, opened 2023-03-03
fix: reset last_run_at on reenabled schedule

lilyli9:
When following the below steps, we expect the datadoc to run at the next scheduled time.
    - Create a schedule, let it run.
    - Disable the schedule
    - Wait for at least one iteration of the schedule (e.g. Hourly schedule, wait more than 1 hour).
    - Re-Enable the schedule and save

However, the actual behavior observed is that the datadoc runs immediately upon reenabling the schedule, instead of waiting for the scheduled time. 

This PR sets the `last_run_at` to the default value (current time) when reenabling a schedule so that the scheduler treats it as a new schedule, ensuring that the time it spent being disabled is not factored into calculations for a datadoc's next run time.

### DIFF
--- a/querybook/server/datasources/datadoc.py
+++ b/querybook/server/datasources/datadoc.py
@@ -9,6 +9,7 @@ from app.auth.permission import (
 from app.datasource import register, api_assert, with_impression
 from app.flask_app import socketio, celery
 from app.db import DBSession, with_session
+from datetime import datetime
 from const.impression import ImpressionItemType
 from const.query_execution import QueryExecutionType
 from env import QuerybookSettings
@@ -327,6 +328,8 @@ def update_datadoc_schedule(id, cron=None, enabled=None, kwargs=None):
             updated_fields["cron"] = cron
         if enabled is not None:
             updated_fields["enabled"] = enabled
+            if enabled and not schedule.enabled:
+                updated_fields["last_run_at"] = datetime.now()
         if kwargs is not None:
             updated_fields["kwargs"] = {
                 **kwargs,


querybook/server/datasources/datadoc.py
```diff
@@ -327,6 +328,8 @@ def update_datadoc_schedule(id, cron=None, enabled=None, kwargs=None):
             updated_fields["cron"] = cron
         if enabled is not None:
             updated_fields["enabled"] = enabled
+            if enabled and not schedule.enabled:
```
czgu, line 331: thanks for the fix! can you add a comment here explaining why it is needed?

lilyli9: Without resetting `last_run_at` whenever a previously running schedule is reenabled, the datadoc runs immediately when it is reenabled instead of the next scheduled time. 

For example, in the following situation:
- `last_run_at` = 3/1/23 at 9am
- Schedule was reenabled at 3/3/23 at 10:15am
- Datadoc was scheduled to run every hour

Based on the current qb logic the scheduler runs periodically to see if it is time to start a run. Since several hours have passed between `last_run_at` and now (the time the schedule was reenabled), and the expected schedule was for the datadoc to run every hour, the datadoc runs as soon as the schedule is enabled again instead of waiting to run at the next expected time. This results in unexpected behavior since in this scenario, the datadoc runs at 3/3/23 at 10:15am instead of 3/3/23 at 11am.

czgu: hey @lilyli9, just to clarify I understand the context on the fix. What I meant is that can you leave a 1 line comment in the code to explain that updating `last_run_at` is needed so it doesn't not retrigger a new run when changing the scheduled datadoc from disabled to enabled

lilyli9: Thanks for the clarification, I added a comment.

PR #1189, opened 2023-03-03
External/disable run doc if already running

lilyli9:
Adds a toggle to the scheduler to allow users to choose whether a scheduled datadoc should run again if the datadoc is already running. This could prevent overlapping runs that would potentially lead to failures or errant population of data. This change references only the most recent run status for the datadoc to determine if the datadoc is still running. 

Screen Shot 2023-03-02 at 11 17 39 AM

### DIFF
--- a/querybook/webapp/components/DataDocSchedule/DataDocScheduleForm.tsx
+++ b/querybook/webapp/components/DataDocSchedule/DataDocScheduleForm.tsx
@@ -85,6 +85,7 @@ const scheduleFormSchema = Yup.object().shape({
                 exporter_params: Yup.object(),
             })
         ),
+        disable_if_running_doc: Yup.boolean().required(),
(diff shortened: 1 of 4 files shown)


querybook/webapp/components/DataDocSchedule/DataDocScheduleForm.tsx
```diff
@@ -215,6 +219,14 @@ export const DataDocScheduleForm: React.FunctionComponent<
                 isValid,
                 dirty,
             }) => {
+                const disableIfRunningDocField = (
+                    <SimpleField
+                        label="Prevent Overlapping Runs"
```
jczhong84, line 224: let's use the positive way to state this as `Allow Concurrent Runs`, make the default as allow to match the current behavior

PR #1191, opened 2023-03-07
feat: add new metadata type - data element

jczhong84:
Add a new metadata type: data element. 

A data element is a basic unit of data that has a specific meaning and can be used to represent information. Data elements are used to represent various types of information, such as names, addresses, dates, numbers, and so on. Each data element has a name or identifier, a data type that specifies the kind of data that can be stored, and a set of properties that determine how the data element can be used. A data element will be associated with a table column. 

This PR is to add the support of loading and displaying data elements from a metastore. Currently we dont support adding a data element through Querybook UI.

image
image


querybook/server/logic/data_element.py
```diff
@@ -0,0 +1,156 @@
...
+                data_element_name=data_element_association.key,
+                column_id=column_id,
+                association_type=data_element_association.type,
+                property_name=DataElementAssociationProperty.KEY,
+                primitive_type=data_element_association.key_primitive,
+                session=session,
+            )
+
+            # add both key and value association only if they are both not None
+            if key_association and value_association:
```
czgu, line 147: value_association is already True here

also why not just add it directly

querybook/server/logic/data_element.py
```diff
@@ -0,0 +1,156 @@
...
+    property_name: DataElementAssociationProperty,
+    primitive_type: str = None,
+    session=None,
+):
+    data_element = None
+    if data_element_name:
+        data_element = get_data_element_by_name(data_element_name, session=session)
+
+    if not data_element and not primitive_type:
+        LOG.error(
```
czgu, line 98: just raise exception here and fail the whole flow with invalid values? Otherwise you might end up with partial data elements

querybook/server/logic/data_element.py
```diff
@@ -0,0 +1,156 @@
...
+    for row in results:
+        resp["type"] = row.type.value
+        resp[row.property_name.value] = (
+            row.data_element.to_dict() if row.data_element else row.primitive_type
+        )
+    return resp
+
+
+@with_session
+def create_or_update_data_element(
```
czgu, line 54: this function is not called anywhere?

querybook/server/models/data_element.py
```diff
@@ -0,0 +1,83 @@
...
+        sql.Integer,
+        sql.ForeignKey("data_table_column.id", ondelete="CASCADE"),
+        nullable=False,
+    )
+    type = sql.Column(
+        sql.Enum(DataElementAssociationType),
+        default=DataElementAssociationType.REF,
+        nullable=False,
+    )
+    property_name = sql.Column(
```
czgu, line 60: make this strings? otherwise we are gonna have fun implementing rows in future if needed lol

querybook/webapp/components/DataElement/DataElement.tsx
```diff
@@ -0,0 +1,102 @@
...
+                        {isDataElement && (
+                            <Icon
+                                name="LayoutGrid"
+                                color="accent"
+                                size={16}
+                                className="mr4"
+                            />
+                        )}
+                        <span>
+                            {typeof dataElement === 'string'
```
czgu, line 36: !isDataElement

querybook/webapp/components/DataElement/DataElementCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+import './DataElement.scss';
+
+interface IProps {
+    dataElement: IDataElement;
+}
+
+export const DataElementCard = ({ dataElement }: IProps) => {
+    const { id, name, type, description, properties } = dataElement;
+
+    const onExternalLinkClick = async () => {
```
czgu, line 18: useCallback

querybook/server/logic/data_element.py
jczhong84: will check the return value of this function to determine if adding both key and value together, will not end up with partial data elements.

The reason of not raising exception to fail the whole flow is, adding data elements is just part of the whole table sync session, would like other parts to be finished even data elements can't be synced succesfully.

querybook/server/logic/data_element.py
jczhong84: not used in this PR, but will be used by the data element sync task

querybook/webapp/components/DataElement/DataElementCard.tsx
jczhong84: As we discussed this before, other than cases like 
 - it will be depended by other hooks
  - it's a true performance issue
 the cost of creating the memoized function itself may not less than the cost of creating the function itself. So i'd prefer not using `useCallback` if not really needed.

czgu (approved):

PR #1192, opened 2023-03-08
fix: add tooltip length

jczhong84:
**Issue**
For long tag tooltips, it is always in single line and will be cut by the table view modal. 

**Fix**
Querybook is using [Balloon.css](https://kazzkiq.github.io/balloon.css/) for tooltips, it acutally supports the length property `data-balloon-length` which can help with this issue.

before:
Snip20230308_8
after:
image


suryabhaskarkarri (commented):
can you add screenshot on the tooltip would appear for large, small lengths ?

jczhong84:
> can you add screenshot of how the tooltip would appear for large, small lengths ?

added two screenshots as before and after.

for other sizes, you may check the official site [Balloon.css](https://kazzkiq.github.io/balloon.css/) to see the differences.

kgopal492 (approved):

PR #1193, opened 2023-03-09
feat: create data element along with syncing table

jczhong84:
When syncing a table, we'll create or update the data element if we find any data elements associated with any columns.

### DIFF
--- a/querybook/server/lib/metastore/base_metastore_loader.py
+++ b/querybook/server/lib/metastore/base_metastore_loader.py
@@ -5,6 +5,7 @@ from typing import Dict, List, Optional, Tuple
 
 import gevent
 from app.db import DBSession, with_session
+from const.data_element import DataElementTuple, DataElementAssociationTuple
 from const.metastore import (
     DataColumn,
     DataOwnerType,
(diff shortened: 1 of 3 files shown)


querybook/server/lib/metastore/base_metastore_loader.py
```diff
@@ -507,6 +534,10 @@ def get_table_and_columns(
         """
         pass
 
+    def get_data_element(self, data_element_name: str) -> DataElementTuple:
+        """Override this to get data element by name"""
+        return None
+
```
suryabhaskarkarri, line 540: is this needed ?  If yes, the return type should be a tuple or none ?

querybook/server/lib/metastore/base_metastore_loader.py
```diff
@@ -534,9 +534,9 @@ def get_table_and_columns(
         """
         pass
 
-    def get_data_element(self, data_element_name: str) -> DataElementTuple:
+    def get_data_element(self, data_element_name: str) -> Optional[DataElementTuple]:
```
suryabhaskarkarri, line 537: @abstractmethod is also needed ?

suryabhaskarkarri (approved):
rest looks good to me.

querybook/server/lib/metastore/base_metastore_loader.py
jczhong84: didn't add AbstractMethod on purpose, for people they dont have data element, they dont have to implement this function.

PR #1194, opened 2023-03-09
fix: mismatched scheduled run times

lilyli9:
- Fixes an issue in [NextRun.tsx](https://github.com/lilyli9/querybook/blob/a97b1564dcc6972d03f07eddf0a5d34959f952d6/querybook/webapp/components/NextRun/NextRun.tsx#L14) to specify that the given cron is based on UTC time and not Local time, and to calculate the next run time accordingly. 

- Also changes "Runs At" to display UTC time instead of Local time to match scheduler, since the selected hour/minute and resultant cron are based on UTC time. Labels to indicate Local/UTC time are also added to the "Runs At" and "Next Run" text for clarification.
Screen Shot 2023-03-08 at 4 11 01 PM

- Added an informational message to the recurrence scheduler page and "Next Run" so that users can check the next run time while they are creating/editing a schedule.
Screen Shot 2023-03-08 at 4 10 21 PM
Screen Shot 2023-03-08 at 4 10 36 PM

### DIFF
--- a/querybook/webapp/components/DataDocScheduleList/DataDocScheduleItem.tsx
+++ b/querybook/webapp/components/DataDocScheduleList/DataDocScheduleItem.tsx
@@ -4,6 +4,7 @@ import React from 'react';
 import { TaskStatusIcon } from 'components/Task/TaskStatusIcon';
 import { formatDuration, generateFormattedDate } from 'lib/utils/datetime';
 import { getWithinEnvUrl } from 'lib/utils/query-string';
+import { cronToRecurrence } from 'lib/utils/cron';
 import { IScheduledDoc } from 'redux/scheduledDataDoc/types';
 import { Link } from 'ui/Link/Link';
 import { AccentText, StyledText, UntitledText } from 'ui/StyledText/StyledText';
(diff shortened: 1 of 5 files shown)


jczhong84:
Thanks for the PR.
 - what you said the bug is on `Next run`, while the change you made is on `Runs at`
 - I think we can have both `Runs at` and `Next Run` to use local time, and it's a good idea to add the trailing `(Local time)`.  
 - Instead of showing `Run Options not Set`, we can just hide it, but only showing it when the schedule is enabled.

baumandm:
@jczhong84 We noticed a number of scenarios where the Runs At is incorrect, like monthly/yearly where the local time ends up being on a different date.  Also it doesn't account for daylight savings time, which can vary depending on where the user is located.

So it seemed like the easiest fix was to leave the Runs At description in UTC, and let Next Run do the local conversions (which it does correctly).

jczhong84:
> @jczhong84 We noticed a number of scenarios where the Runs At is incorrect, like monthly/yearly where the local time ends up being on a different date. Also it doesn't account for daylight savings time, which can vary depending on where the user is located.
> 
> So it seemed like the easiest fix was to leave the Runs At description in UTC, and let Next Run do the local conversions (which it does correctly).

thanks for the context!

put some comments and please resolve the build failure

PR #1195, opened 2023-03-10
feat: display data element description as column description

jczhong84:
- when there is no column description, we'll show the associated data element's description.
- make the table property clickable if it's a link

**with column description**
Snip20230310_17

**without column description**
Snip20230310_14
Snip20230310_13

### DIFF
--- a/querybook/webapp/components/DataElement/DataElementDescription.tsx
+++ b/querybook/webapp/components/DataElement/DataElementDescription.tsx
@@ -1,0 +1,69 @@
+import React from 'react';
+
+import {
+    DataElementAssociationType,
+    IDataElementAssociation,
+} from 'const/dataElement';
+import { KeyContentDisplay } from 'ui/KeyContentDisplay/KeyContentDisplay';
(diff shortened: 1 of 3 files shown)


querybook/webapp/components/DataElement/DataElementDescription.tsx
```diff
@@ -0,0 +1,69 @@
...
+        </div>
+    );
+
+    const listDescriptionDOM = (
+        <div>
+            <KeyContentDisplay keyString={valueName}>
+                <AccentText>{valueDescription}</AccentText>
+            </KeyContentDisplay>
+        </div>
+    );
```
suryabhaskarkarri, line 51: may be its not necessary to set these constants always.   Can this be a function and return appropriate response depending on type ?
For ex:  majority will be non-composite elements.  So its not necessary to always define and compute, map & list description DOM's


const descriptionDom = (type) => {
      if noncomposite, return <div>
      if array -> just get valueDataElement.description
      etc ?
}

suryabhaskarkarri (approved):

PR #1199, opened 2023-03-17
fix: some bug fix related to tag and data element

jczhong84:
some bug fixes
 - can't add tags whose name contains space and underscore
 - too many api calls on the column tab in table view. Combined the tag, data element and column stats into one api call
 - unnecessary wrap of the tag name on the left sidebar mini table view. Change it to nowrap and also hide tag type in the mini view.
 
Also removed the expand/collapse toggle button and move the column name before the column type
image

### DIFF
--- a/querybook/server/logic/data_element.py
+++ b/querybook/server/logic/data_element.py
@@ -48,7 +48,7 @@ def get_data_element_association_by_column_id(
     for row in associations:
         data_element["type"] = row.type.value
         data_element[row.property_name] = (
-            row.data_element.to_dict() if row.data_element else row.primitive_type
+            row.data_element if row.data_element else row.primitive_type
         )
     return data_element
 
--- a/querybook/server/logic/data_element.py
+++ b/querybook/server/logic/data_element.py
@@ -1,6 +1,7 @@
 from typing import Union
 from app.db import with_session
 from const.data_element import (
+    DataElementAssociationDict,
     DataElementAssociationProperty,
     DataElementAssociationTuple,
     DataElementAssociationType,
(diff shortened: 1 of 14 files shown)


querybook/server/logic/data_element.py
```diff
@@ -24,12 +25,32 @@ def get_data_element_by_name(name: str, session=None):
...
+        LOG.error(
+            f"Column {column_id} has more than one data element associated with it"
+        )
+        return None
+
+    data_element = {}
+    for row in associations:
+        data_element["type"] = row.type.value
+        data_element[row.property_name] = (
+            row.data_element.to_dict() if row.data_element else row.primitive_type
```
czgu, line 51: don't serialize it here. Let it be serialized automatically when returning

querybook/server/logic/data_element.py
```diff
@@ -75,7 +96,11 @@ def create_data_element_association(
     primitive_type: str = None,
     session=None,
 ):
-    if data_element_tuple is None:
+    if (
+        data_element_tuple is None
+        or not data_element_tuple.name
```
czgu, line 101: nit: how about if not data_element_tuple or not data_element_tuple.name ...
so its more consistent to read

querybook/webapp/components/DataTableStats/DataTableColumnStats.tsx
```diff
@@ -1,27 +1,19 @@
...
 interface IProps {
-    columnId: number;
+    stats: ITableColumnStats[];
 }
 
 export const DataTableColumnStats: React.FunctionComponent<IProps> = ({
-    columnId,
+    stats,
 }) => {
-    const { data: tableColumnStats } = useResource(
```
czgu, line 17: since we don't get tags, data elements, and stats individually anymore, can you remove the api and resource?

querybook/webapp/const/metastore.ts
```diff
@@ -116,6 +119,9 @@ export interface IDataColumn {
     name: string;
     table_id: number;
     type: string;
+    stats?: ITableColumnStats[];
```
czgu, line 122: instead of stats?, can you separate the type out to be IDetailedDataColumn extends IDataColumn?

querybook/server/logic/data_element.py
jczhong84: api json encoder wont do automatic serialization for nested objects.

querybook/webapp/components/DataTableStats/DataTableColumnStats.tsx
jczhong84: actually there are 3 `/column/stats` apis, two of them are also not used today. should we remove all of them?

czgu (approved):

querybook/webapp/components/DataTableStats/DataTableColumnStats.tsx
czgu: oh yeah lets remove all of them then

czgu (approved):

PR #1204, opened 2023-03-25
feat: add stats logging

jczhong84:
Adding stats logging support. 

Metrics:
  - Number of active users
  - Number of API requests
  - Latency of API requests
  - Number of websocket connections
  - Number of sql session failures
  - Number of scheduled system task failures
  - Number of scheduled datadoc failures
  - Latency of Redis operations
  - Number of query executions


querybook/server/app/auth/__init__.py
```diff
@@ -1,14 +1,32 @@
...
 from const.path import BUILD_PATH
 from const.datasources import DS_PATH
 from env import QuerybookSettings
 from lib.utils.import_helper import import_module_with_default
+from lib.stats_logger import stats_logger, ACTIVE_USER_COUNTER
 
 
 auth = None
 login_config = None
+active_users = {}
```
czgu, line 15: this wouldn't work if you have multiple servers, unless you plan to store this stat in redis

querybook/server/app/auth/__init__.py
```diff
@@ -1,14 +1,32 @@
...
+
+
+def update_active_users():
+    if current_user.id not in active_users:
+        stats_logger.incr(ACTIVE_USER_COUNTER)
+
+    active_users[current_user.id] = time.time()
+
+    for user_id, last_activity_time in list(active_users.items()):
+        if time.time() - last_activity_time > ACTIVE_USER_SESSION_TIMEOUT:
```
czgu, line 27: maybe we don't need to log the num of active users? 3 reasons:
1. it is quite annoying to handle with multiple servers
2. the stat is not as useful as number of requests which we also have
3. the num of user can also be extracted from impression log
4. this call is not real time, which pads a variable processing time to the request

querybook/server/app/datasource.py
```diff
@@ -56,6 +61,11 @@ def wrapper(fn):
         @flask_app.route(r"%s%s" % (DS_PATH, url), methods=methods)
         @functools.wraps(fn)
         def handler(**kwargs):
+            # increment the number of api request counter
+            stats_logger.incr(API_REQUEST_COUNTER.format(fn.__name__))
```
czgu, line 65: should we use the url here? it is generic right

querybook/server/datasources_socketio/helper.py
```diff
@@ -18,7 +19,13 @@ def handler(*args, **kwargs):
             if not current_user.is_authenticated:
                 LOG.error("Unauthorized websocket access")
                 disconnect()
+                # decrement ws connections counter on disconnect
+                stats_logger.decr(WS_CONNECTIONS_COUNTER)
             else:
+                # decrement ws connections counter on disconnect
+                if url == "disconnect":
+                    stats_logger.decr(WS_CONNECTIONS_COUNTER)
```
czgu, line 27: do the logging after disconnect function is done?

querybook/server/tasks/all_tasks.py
```diff
@@ -54,3 +59,12 @@ def configure_workers(sender=None, conf=None, **kwargs):
         clean_up_query_execution()
     else:
         LOG.info(f"Starting DEV Celery worker: {sender}")
+
+
+@task_failure.connect
+def handle_task_failure(sender, signal, *args, **kwargs):
+    task_name = sender.name
+    if task_name.startswith("tasks.run_datadoc"):
```
czgu, line 67: there is a task_type field for task and it is either "prod" or "user". Can you make the two connect?

querybook/server/app/datasource.py
jczhong84: url is something like `/query_execution/<int:query_execution_id>/`, which will include the parameter types.

czgu (approved):

PR #1207, opened 2023-03-31
feat: add configurable timeout to querybook

czgu:
- added force terminate when cancelling queries
- added timeout config for presto
- make sure numberfield always return a number


querybook/server/lib/query_executor/clients/presto.py
```diff
@@ -11,6 +11,8 @@ def __init__(self, cursor: presto.Cursor) -> None:
         self._cursor = cursor
         self._init_query_state_vars()
 
+        print(cursor._requests_kwargs)
```
jczhong84, line 14: remove debug print

jczhong84 (approved):

PR #1210, opened 2023-04-04
feat: Fix Editor endpoint.

praful-ashirwad:
Fix Editor Function at Editor Endpoint : /datadoc/<int:doc_id>/editor/<int:uid>/
and Web EndPoint -> http://localhost:10001/demo_environment/datadoc/2/doc-3/

BUG:
If we share private datadoc ( in read-only) to user's and after that.. if they ask for  'edit' the datadoc.. owner can't give them permission, because it's already created in DataDocEditor.. So, we need to call "update_datadoc_editor".. instead of this.. i created "create_or_update_datadoc_edtor" function and it's fix.

### DIFF
--- a/querybook/server/datasources/datadoc.py
+++ b/querybook/server/datasources/datadoc.py
@@ -442,9 +442,16 @@ def add_datadoc_editor(
 ):
     with DBSession() as session:
         assert_can_write(doc_id, session=session)
-        editor = logic.create_data_doc_editor(
-            data_doc_id=doc_id, uid=uid, read=read, write=write, commit=False
-        )
+
+        # CREATE & UPDATE FUNCTION ADDED
+        editor = logic.create_or_update_data_doc_editor(data_doc_id=doc_id, uid=uid, read=read, write=write, commit=False)
+        
+        # ONLY CREATE FUNCTION COMMENT
+        #
+        # editor = logic.create_data_doc_editor(
+        #     data_doc_id=doc_id, uid=uid, read=read, write=write, commit=False
+        # )
+        
         editor_dict = editor.to_dict()
 
         access_request = logic.get_data_doc_access_request_by_doc_id(
--- a/querybook/server/logic/datadoc.py
+++ b/querybook/server/logic/datadoc.py
@@ -860,6 +860,39 @@ def update_data_doc_editor(
             session.refresh(editor)
         return editor
 
+@with_session
+def create_or_update_data_doc_editor(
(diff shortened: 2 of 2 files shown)


querybook/server/datasources/datadoc.py
```diff
@@ -442,9 +442,16 @@ def add_datadoc_editor(
...
-        )
+
+        # CREATE & UPDATE FUNCTION ADDED
+        editor = logic.create_or_update_data_doc_editor(data_doc_id=doc_id, uid=uid, read=read, write=write, commit=False)
+        
+        # ONLY CREATE FUNCTION COMMENT
+        #
+        # editor = logic.create_data_doc_editor(
+        #     data_doc_id=doc_id, uid=uid, read=read, write=write, commit=False
+        # )
```
jczhong84, line 453: please remove these comments

querybook/server/datasources/datadoc.py
```diff
@@ -442,9 +442,16 @@ def add_datadoc_editor(
 ):
     with DBSession() as session:
         assert_can_write(doc_id, session=session)
-        editor = logic.create_data_doc_editor(
-            data_doc_id=doc_id, uid=uid, read=read, write=write, commit=False
-        )
+
+        # CREATE & UPDATE FUNCTION ADDED
```
jczhong84, line 446: Please update this comment to be more meaningful or just remove it, as `create_or_update_data_doc_editor` is quite self-explanatory

querybook/server/logic/datadoc.py
```diff
@@ -813,6 +813,15 @@ def get_data_doc_editor_by_id(id, session=None):
 def get_data_doc_editors_by_doc_id(data_doc_id, session=None):
     return session.query(DataDocEditor).filter_by(data_doc_id=data_doc_id).all()
 
+# Return Instance of unique data
```
jczhong84, line 816: use docstring `"""..."""` instead, or just remove it as this function is also quite self-explanatory

querybook/server/datasources/datadoc.py
```diff
@@ -442,9 +442,16 @@ def add_datadoc_editor(
...
+        editor = logic.create_or_update_data_doc_editor(data_doc_id=doc_id, uid=uid, read=read, write=write, commit=False)
```
jczhong84, line 447: this function name `create_or_update_data_doc_editor` doesn't match the function name below, which doesn't exist

jczhong84 (commented):
Thanks for finding and fixing the issue.

PR #1212, opened 2023-04-05
specify module versions to avoid errors during launch in devapp 

wcypig:
not sure if others have this issue as well, when calling make for the first time, getting errors due to module version incompatibility. for example:

EntryPoints' object has no attribute 'get', which requires setting importlib-metadata below version 5.0

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,7 +1,11 @@
 ipython
 watchdog_gevent
 watchdog[watchmedo]
-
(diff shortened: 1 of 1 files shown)


jczhong84:
@wcypig Thanks for the PR. 

I tried to run on my local machine, not seeing the issues. You can put those into your `local.txt` while we can wait and see if there are more people reporting the same issue.

PR #1217, opened 2023-04-07
fix: skip empty query cells when run datadoc

jczhong84:
scheduled datadoc will fail when there is a empty query cell in the doc. Here we skip empty cells.


querybook/server/tasks/run_datadoc.py
```diff
@@ -73,10 +73,15 @@ def run_datadoc_with_config(
         # Prepping chain jobs each unit is a [make_qe_task, run_query_task] combo
         for index, query_cell in enumerate(query_cells):
             engine_id = query_cell.meta["engine"]
+            raw_query = query_cell.context
+
+            # Skip empty cells
+            if not raw_query or raw_query.isspace():
```
czgu, line 79: is context nullable? why not `if len(raw_query.strip()) == 0`

czgu (approved):

PR #1221, opened 2023-04-07
fix: filestore performance fix

najma-kunju:


querybook/server/lib/result_store/stores/file_store.py
```diff
@@ -28,10 +28,11 @@ def __init__(self, uri: str):
 
     def start(self):
         self._chunks_length = 0
+        self._chunks = []
         os.makedirs(self.uri_dir_path, exist_ok=True)
 
     def write(self, data: str):
-        # write each line into csv
+        # Append each line to a list and then dump at once
```
jczhong84, line 35: dumping at once may cause issue when the result set size to be uploaded is too big. 
A better way could be dumping by chunks, collecting X rows and then dump it.

jczhong84 (commented):
Thanks for the PR!

could you also add some descriptions to state what the issue is and what improvement you made?

querybook/server/lib/result_store/stores/file_store.py
najma-kunju: About my PR:

Previously, the write method in FileUploader class was opening the file in append mode and writing data into it for each record of the result set of the query. Which perhaps was talking a lot of time if the result set is huge and is also not a good practice to increase the file operation. 

So to increase the performance, we did the below:
-        Initialised self._chunks as an empty list in start method of FileUploader class.
-        Kept appending each record of the query result to the above list in write method of FileUploader class. 
-        At the end of the process, Converted the list elements to string using join in end method of FileUploader class.
-        Opened the file in write mode and wrote the above string value to file in end method of FileUploader class.

This was tested and we observed that the time taken to write the data to file was in a few seconds.

I appreciate you in suggesting an alternative approach for dumping the data. 

Based on the approach as explained above is there still a need of dumping by chunks as we are using join iterator to convert list to string and then write that to file. 

Please feel free to correct me if I am wrong.

Thanks for your time.

najma-kunju:
About my PR:

Previously, the write method in FileUploader class was opening the file in append mode and writing data into it for each record of the result set of the query. Which perhaps was talking a lot of time if the result set is huge and is also not a good practice to increase the file operation.

So to increase the performance, we did the below:

-    Initialised self._chunks as an empty list in start method of FileUploader class.
-    Kept appending each record of the query result to the above list in write method of FileUploader class. 
-    At the end of the process, Converted the list elements to string using join in end method of FileUploader class.
-    Opened the file in write mode and wrote the above string value to file in end method of FileUploader class.

Thanks for your time.

querybook/server/lib/result_store/stores/file_store.py
jczhong84: thanks for the for detailed explanation.
yes, there is no doubt here that the current way of opening the file and append it for each record doesn't have good performance. 
What I meant was for the extreme case like:
when the file to be uploaded is too big to be fit into the container's allocated memory, you cant just keep appending each row to `self._chunks` and keep it in the memory, but have to write it the file in batches.

querybook/server/lib/result_store/stores/file_store.py
```diff
@@ -27,25 +27,43 @@ def __init__(self, uri: str):
...
+
+    def unload(self):
+        data = "".join(self._chunks)
         with open(self.uri, "a") as result_file:
             result_file.write(data)
-        return True
 
     def end(self):
-        pass
+        if len(self._chunks) < self._chunk_size:
```
jczhong84, line 65: this checks seem unnecessary to me, as it should not happen

querybook/server/lib/result_store/stores/file_store.py
```diff
@@ -27,25 +27,43 @@ def __init__(self, uri: str):
...
             and self._chunks_length + data_len > QuerybookSettings.DB_MAX_UPLOAD_SIZE
         ):
             return False
-
+        
         self._chunks_length += data_len
+        # with open(self.uri, "a") as result_file:
+        #     result_file.write(data)
+
+        if len(self._chunks) == self._chunk_size:
```
jczhong84, line 52: move the check and appending below the line
```
self._chunks.append(data)
```

querybook/server/lib/result_store/stores/file_store.py
```diff
@@ -27,25 +27,43 @@ def __init__(self, uri: str):
...
 
     def start(self):
+        self._chunk_size=100000
         self._chunks_length = 0
+        self._chunks = []
         os.makedirs(self.uri_dir_path, exist_ok=True)
 
+    def _reset_variables(self):
+        self._chunks = []
+        self._chunks_length = 0
```
jczhong84, line 37: also `_chunks_length` should not be reset as 0, as it is used to compare with `QuerybookSettings.DB_MAX_UPLOAD_SIZE` , it needs to be the total length.

querybook/server/lib/result_store/stores/file_store.py
```diff
@@ -27,25 +27,43 @@ def __init__(self, uri: str):
         self.uri = get_file_uri(uri)
 
     def start(self):
+        self._chunk_size=100000
```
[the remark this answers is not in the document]
najma-kunju: updated

querybook/server/lib/result_store/stores/file_store.py
```diff
@@ -27,25 +27,43 @@ def __init__(self, uri: str):
...
         self._chunks_length = 0
+        self._chunks = []
         os.makedirs(self.uri_dir_path, exist_ok=True)
 
+    def _reset_variables(self):
```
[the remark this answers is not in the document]
najma-kunju: updated

querybook/server/lib/result_store/stores/file_store.py
```diff
@@ -27,25 +27,43 @@ def __init__(self, uri: str):
...
         if (
             QuerybookSettings.DB_MAX_UPLOAD_SIZE > 0
             and self._chunks_length + data_len > QuerybookSettings.DB_MAX_UPLOAD_SIZE
         ):
             return False
-
+        
         self._chunks_length += data_len
+        # with open(self.uri, "a") as result_file:
+        #     result_file.write(data)
```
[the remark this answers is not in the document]
najma-kunju: updated

querybook/server/lib/result_store/stores/file_store.py
najma-kunju: updated

querybook/server/lib/result_store/stores/file_store.py
najma-kunju: updated

querybook/server/lib/result_store/stores/file_store.py
najma-kunju: This check would be used when the length of list appended (self._chunks) is less than the batch size defined as env variable. 
ie, if the number of rows retrieved from query is less than env variable or after making batches, the length of the last  batch would be less than the env variable.

This is tested and is working as expected.

querybook/server/lib/result_store/stores/file_store.py
```diff
@@ -27,25 +27,38 @@ def __init__(self, uri: str):
...
             and self._chunks_length + data_len > QuerybookSettings.DB_MAX_UPLOAD_SIZE
         ):
             return False
-
+        
         self._chunks_length += data_len
+        
+        self._chunks.append(data)
+
+        if len(self._chunks) == self._chunk_size:
```
jczhong84, line 48: Just realized that the `DB_MAX_UPLOAD_SIZE` is the actual data size, while this `_chunk_size` is the number of records. we may want to make the chunk size based on the data size as well (multiples of 1024)

najma-kunju: Hi @jczhong84,
Thanks for reviewing my fix. 

A question that popped in my head is , below check remains valid for dbstore as we are making entry to database. In filestore, as we are writing to file , Do we need the same? 
if (
            QuerybookSettings.DB_MAX_UPLOAD_SIZE > 0
            and self._chunks_length + data_len > QuerybookSettings.DB_MAX_UPLOAD_SIZE
        ):
            return False

querybook/server/lib/result_store/stores/file_store.py
najma-kunju: updated the same.

querybook/server/lib/result_store/stores/file_store.py
jczhong84: thanks for pointing this out. 
could you add a separate max size setting for the file store? 
can set a big number for it as default, or just set 0 as unlimited by default.

najma-kunju: hi @jczhong84,
Defining the max size of file to 0 does not work because even if the number of rows returned from the query is 0, it will anyhow write the column names to the file. 

But if we are setting it to a high value,  Im wondering does it serve the purpose of choosing filestore above dbstore? Correct me if im wrong.

jczhong84: if the query result has 0 rows, isn't it expected to write a empty file with only the column names?
e.g. set DB_MAX_UPLOAD_SIZE as 0, it will bypass below check, which basically means no limit check.
```
        if (
            QuerybookSettings.DB_MAX_UPLOAD_SIZE > 0
            and self._chunks_length + data_len > QuerybookSettings.DB_MAX_UPLOAD_SIZE
        ):
            return False
```

what matters is the file size, not the number of rows. what I meant is  by default set it as 0 (unlimited), but people can set their own limit like 10GB to override this default setting

najma-kunju: Hi @jczhong84,
Have made the amendments as suggested.

Regards,
Najma

jczhong84: thanks for the update.

use 
`if _chunks_length >= self._chunk_size:`
but not
`if len(self._chunks) == self._chunk_size:`

najma-kunju: hi @jczhong84 ,
 have changed the code and logic to use self._chunks_length.

jczhong84:
@najma-kunju there are still lint and python test failures

najma-kunju:
@jczhong84, have fixed the current 2 lint errors

jczhong84:
@najma-kunju it's still failing

najma-kunju:
Hi @jczhong84, 
Now the checks have passed. 
Thanks for the help throughout PRs 1221 and 1237.

querybook/tests/test_lib/test_result_store/test_stores/test_file_store.py
```diff
@@ -72,9 +72,9 @@ def mock_write_file(s: str):
             uploader.write("foo,bar,baz\n")
             uploader.write('"hello world", "foo\nbar", ","\n')
 
-            uploader.end()
-
-        m.assert_called_with(f"{FILE_STORE_PATH}test/path", "a")
+            # uploader.end()
+        mock_file_content = 'foo,bar,baz\n"hello world", "foo\nbar", ","\n'
+        # m.assert_called_with("test/path", "a")
```
jczhong84, line 77: why commenting here and above two `uploader.end()` out?

PR #1226, opened 2023-04-11
feat: Fix Data Doc Editor

praful-ashirwad:
#1210 -> please refer as same.

Due to some problem.. i recreate this pull request.

Fix Data Doc Editor.. In which Owner give private Data Doc permission to users, but problem was.. when we add user and user ask for edit permission.. we got error "Duplicate Key", Bcoz we add it already.. so create function all... and so, i create new function "create_or_update_data_doc_editor" and comments i added.

Thanx


czgu:
thanks for contributing to Querybook! 
the code lgtm, once it pass the black auto formatting we can merge

PR #1231, opened 2023-04-19
fix: LIMIT ALL can now be used in Querybook

jij1949:
Trino supports [`LIMIT ALL`](https://trino.io/docs/current/sql/select.html#limit-or-fetch-first-clause) but Querybook does not – it automatically adds an additional limit which breaks.

This change disables the automatic limiter when LIMIT ALL is used.

### DIFF
--- a/querybook/webapp/lib/sql-helper/sql-limiter.ts
+++ b/querybook/webapp/lib/sql-helper/sql-limiter.ts
@@ -118,8 +118,8 @@ export function getLimitedQuery(
         .map((statement) => {
             const existingLimit = getSelectStatementLimit(statement, language);
             if (
-                existingLimit == null || 
-                existingLimit >= 0 || 
+                existingLimit == null ||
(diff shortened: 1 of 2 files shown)


querybook/webapp/lib/sql-helper/sql-limiter.ts
```diff
@@ -45,13 +45,22 @@ export function getSelectStatementLimit(
...
+        return Number(matchLimitPatternNum[1].text);
+    }
+
+    const matchLimitPatternAll = tokenPatternMatch(outerStatement, [
+        { type: 'KEYWORD', text: 'limit' },
+        { type: 'KEYWORD', text: 'all' },
+    ]);
+
+    if (matchLimitPatternAll) {
+        return -2;
```
jczhong84, line 63: please update the documentation of function accordingly.

jij1949:
> @jij1949 could you please fixed the lint errors?

PR #1239, opened 2023-05-01
Improve documentation for some editor-related functions

adamobeng:
- explain the function and output format of `getEditorLines`
- explain the contents of `Line` class
- describe how `codeAnalysis` is used

### DIFF
--- a/querybook/webapp/lib/sql-helper/sql-lexer.ts
+++ b/querybook/webapp/lib/sql-helper/sql-lexer.ts
@@ -827,34 +827,34 @@ export function getEditorLines(statements: IToken[][]): Line[] {
     //      FROM main.products;
     //
     // And return value would be:
-    [
-        { // line 0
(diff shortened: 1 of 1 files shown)


querybook/webapp/lib/sql-helper/sql-lexer.ts
```diff
@@ -809,7 +815,48 @@ export function findTableReferenceAndAlias(statements: IToken[][]) {
...
+        { // line 2
+            "statements": [[0, 1]], // statement 2 of the query starts at character 0
+            "contexts": [
+                [0, "column"], // char 0-4: column name
+                [5, "table"], // char 5-19: table name
+                [19, "none"] // char 19-end: neither
+            ]
+        }
+    ]
+
```
jczhong84, line 858: this needs to be inside the comment I guess? also can you add this whole comments above the function by using
/**
*/

PR #1245, opened 2023-05-10
ui: comment db updates

meowcodes:
meeting requirements below

Support comment and comment threads (1 level)
Each comment can be edited after posted
Each comment can be attached to a datadoc cell, or a whole table (make sure this system is extensible if we want to add commenting elsewhere)
When a cell is cloned or copied, the comment is not copied. When a cell is cut, the comment is moved
Support reactions, like Slack
Support richtext in comments

### DIFF
--- a/querybook/server/models/comment.py
+++ b/querybook/server/models/comment.py
@@ -1,0 +1,108 @@
+import sqlalchemy as sql
+from app import db
+from lib.sqlalchemy import CRUDMixin
+from sqlalchemy.orm import relationship
+from const.db import mediumtext_length, name_length, now
+
+Base = db.Base
+
+
+class Comment(CRUDMixin, Base):
+    __tablename__ = "comment"
+
+    id = sql.Column(sql.Integer, primary_key=True)
+
+    created_at = sql.Column(sql.DateTime, default=now)
+    updated_at = sql.Column(sql.DateTime, default=now)
+    created_by = sql.Column(
+        sql.Integer, sql.ForeignKey("user.id", ondelete="SET NULL"), nullable=True
+    )
+
+    text = sql.Column(sql.Text(length=mediumtext_length))
+
+    has_thread = sql.Column(sql.Boolean, default=False)
+
+
+class CommentReaction(CRUDMixin, Base):
+    __tablename__ = "comment_reaction"
+
+    id = sql.Column(sql.Integer, primary_key=True, autoincrement=True)
+    comment_id = sql.Column(
+        sql.Integer,
+        sql.ForeignKey("comment.id", ondelete="CASCADE"),
+        index=True,
+        nullable=False,
+        unique=True,
+    )
+    reaction = sql.Column(
+        sql.String(length=name_length), index=True, unique=True, nullable=False
+    )
+    created_by = sql.Column(
+        sql.Integer, sql.ForeignKey("user.id", ondelete="SET NULL"), nullable=True
+    )
+
+
+class DataTableComment(CRUDMixin, Base):
+    __tablename__ = "data_table_comment"
+
+    id = sql.Column(sql.Integer, primary_key=True, autoincrement=True)
+    data_table_id = sql.Column(
+        sql.Integer,
+        sql.ForeignKey("data_table.id", ondelete="CASCADE"),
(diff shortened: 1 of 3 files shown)


querybook/server/models/comment.py
```diff
@@ -0,0 +1,108 @@
...
+        sql.ForeignKey("comment.id", ondelete="CASCADE"),
+        nullable=False,
+        unique=True,
+    )
+
+    data_cell = relationship("DataCell", uselist=False)
+    comment = relationship("Comment", uselist=False)
+
+
+class NestedComment(CRUDMixin, Base):
```
jczhong84, line 89: It seems this tables is not necessary to me. we can add this `parent_comment_id` to the `comment` table directly and `child_comment_id` is actually not needed.

jczhong84:
also, do we need to introduce the concept of `resolve a comment`? and can a datacell have multiple comments?

querybook/server/models/comment.py
```diff
@@ -0,0 +1,108 @@
...
+    id = sql.Column(sql.Integer, primary_key=True, autoincrement=True)
+    comment_id = sql.Column(
+        sql.Integer,
+        sql.ForeignKey("comment.id", ondelete="CASCADE"),
+        index=True,
+        nullable=False,
+        unique=True,
+    )
+    reaction = sql.Column(
+        sql.String(length=name_length), index=True, unique=True, nullable=False
```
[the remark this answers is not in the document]
meowcodes: it needs to be indexed so i can get the count easily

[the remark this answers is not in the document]
jczhong84: as the count is always for a specific comment and `comment_id` is already indexed, and also there wont be too many reactions for a comment, so I dont think it's necessary to index it.

querybook/server/models/comment.py
```diff
@@ -0,0 +1,84 @@
...
+    created_by = sql.Column(
+        sql.Integer, sql.ForeignKey("user.id", ondelete="SET NULL"), nullable=True
+    )
+
+    text = sql.Column(sql.Text(length=mediumtext_length))
+
+    parent_comment_id = sql.Column(
+        sql.Integer,
+        sql.ForeignKey("comment.id", ondelete="CASCADE"),
+        index=True,
```
czgu, line 26: why are we indexing on foreign keys? https://stackoverflow.com/questions/304317/does-mysql-index-foreign-key-columns-automatically

querybook/server/models/comment.py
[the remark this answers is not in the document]
jczhong84: to clarify, what meant is keep `index=True` for `comment_id`, while remove `index=True` for `reaction`

querybook/server/models/comment.py
```diff
@@ -0,0 +1,80 @@
...
+class CommentReaction(CRUDMixin, Base):
+    __tablename__ = "comment_reaction"
+
+    id = sql.Column(sql.Integer, primary_key=True, autoincrement=True)
+    comment_id = sql.Column(
+        sql.Integer,
+        sql.ForeignKey("comment.id", ondelete="CASCADE"),
+        nullable=False,
+    )
+    reaction = sql.Column(sql.String(length=name_length), index=True, nullable=False)
```
czgu, line 39: unless we are filtering on reaction, there is no point for an index? we would always filter it by comment id right?

querybook/server/models/comment.py
```diff
@@ -0,0 +1,80 @@
...
+        sql.ForeignKey("data_table.id", ondelete="CASCADE"),
+        nullable=False,
+    )
+    comment_id = sql.Column(
+        sql.Integer,
+        sql.ForeignKey("comment.id", ondelete="CASCADE"),
+        nullable=False,
+    )
+
+    data_table = relationship("DataTable", uselist=False)
```
czgu, line 60: nit: remove these 2 relationships (i dont think this table will be used directly anyways)

querybook/server/models/comment.py
meowcodes: i guess we can aggregate reaction count on the front end

czgu (approved):

PR #1246, opened 2023-05-10
ui: long table name fix

meowcodes:
Screenshot 2023-05-10 at 2 54 24 PM
Screenshot 2023-05-10 at 2 54 06 PM
Screenshot 2023-05-10 at 2 53 44 PM
Screenshot 2023-05-10 at 2 53 04 PM

wrapping long table names/any search res title

### DIFF
--- a/querybook/webapp/components/Search/SearchResultItem.scss
+++ b/querybook/webapp/components/Search/SearchResultItem.scss
@@ -31,15 +31,8 @@
                 padding-right: var(--padding-sm);
                 align-items: baseline;
 
-                > div:first-child {
-                    flex-grow: 1;
-                    overflow: hidden;
-                    .Text {
-                        width: 100%;
-                        .result-item-title {
-                            overflow-wrap: break-word;
-                        }
-                    }
+                .result-item-title {
+                    overflow-wrap: anywhere;
                 }
 
                 .result-item-timestamp {
--- a/querybook/webapp/components/Search/SearchResultItem.scss
+++ b/querybook/webapp/components/Search/SearchResultItem.scss
@@ -31,10 +31,18 @@
                 padding-right: var(--padding-sm);
                 align-items: baseline;
 
-                .result-item-title {
-                    overflow: hidden;
+                > div:first-child {
+                    overflow-wrap: break-word;
+                    max-width: calc(100% - 140px);
+                    div {
+                        width: 100%;
+                        text-align: left;
+                    }
+                }
+                > div:last-child {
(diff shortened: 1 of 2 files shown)


querybook/webapp/components/Search/SearchResultItem.scss
```diff
@@ -30,11 +30,23 @@
...
+                    flex-grow: 1;
                     overflow: hidden;
+                    .Text {
+                        width: 100%;
+                        .result-item-title {
+                            overflow-wrap: break-word;
+                        }
+                    }
+                }
+
```
jczhong84, line 45: for the title, I think you only need
```
.result-item-title {
    overflow-wrap: anywhere
}
```
`break-word` doesn't work on underscores

querybook/webapp/components/Search/SearchResultItem.scss
```diff
@@ -30,11 +30,23 @@
             .result-items-top {
                 padding-right: var(--padding-sm);
                 align-items: baseline;
+                width: 100%;
```
jczhong84, line 33: dont think this is needed

querybook/webapp/components/Search/SearchResultItem.scss
```diff
@@ -30,11 +30,23 @@
...
                     overflow: hidden;
+                    .Text {
+                        width: 100%;
+                        .result-item-title {
+                            overflow-wrap: break-word;
+                        }
+                    }
+                }
+
+                > div:last-child:not(:first-child) {
```
jczhong84, line 46: for the timestamp, could you add a class name for it, instead of using last-child?

and I think you only need something like, without min-width
```
.result-item-timestamp {
  flex-shrink: 0
}
```

meowcodes: Screenshot 2023-05-12 at 1 41 05 PM

querybook/webapp/components/Search/SearchResultItem.scss
meowcodes: Screenshot 2023-05-12 at 1 46 20 PM

neither does what you're suggesting. it looks the same as the current implementation

querybook/webapp/components/Search/SearchResultItem.scss
meowcodes: also without min-width, the length of the title will vary based on the length of the timestamp, which looks messy

jczhong84 (approved):

Issue #1259, opened 2023-05-31
Injectable Secrets into the Helm Chart


hawkesn:
The current design of the querybook helm chart does not allow you to use an existing secret if it were created by external resources.
Tooling such as:
- [external-secrets](https://external-secrets.io/v0.8.3/)
- [sealed-secrets](https://github.com/bitnami-labs/sealed-secrets)
- [sops-secret-operator](https://github.com/isindir/sops-secrets-operator)
- [vault](https://github.com/hashicorp/vault-k8s)

Handles best practices around secret management and allows you to safely inject secrets into applications without having them be stored as plain text in your `values.yaml`

Allowing a value to use an existing secret, formatted equivalently as the current secret, would be ideal.

Here's an example from the `datadog` chart:
Option to specify the secret, or an existing secret: https://github.com/DataDog/helm-charts/blob/main/charts/datadog/values.yaml#L29-L49
How to handle if the `existingSecret` value is populated: https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/secret-api-key.yaml

PR #1260, opened 2023-06-01
feat: Injectable Secrets into the Helm Chart

hawkesn:
Fixes: https://github.com/pinterest/querybook/issues/1259
Allows a user to specify the name of an existing Kubernetes Secret, or defaults to the secret specified in `values.yaml`


jczhong84:
Thanks for the fix! As we're not using Kubernetes/Helm, we'll just trust you on this. Please make sure you've tested on your environment properly.

hawkesn:
Hi @jczhong84 , thanks! 
Yes, we are actively using this helm chart change in our production environment, it works well for our use case.
Let me know if there's anything I can do to get an approval.

jczhong84 (approved):

PR #1262, opened 2023-06-05
backend: add comment endpoints

meowcodes:
QQQs
- should i add permission checks for adding comments?
- should users be allowed to delete comments? should they be completely deleted or archived (and say deleted comment or something like that in UI)
- did i set up the model correctly so that if a Comment was deleted, its corresponding DataTableComment or DataCellComment would be deleted? i think i did but just wanna confirm.

### DIFF
--- a/querybook/server/const/comment.py
+++ b/querybook/server/const/comment.py
@@ -1,0 +1,15 @@
+from typing import List, Optional, TypedDict
+
+from querybook.server.models.comment import Comment, CommentReaction
+
+
+class CommentDict(TypedDict):
+    id: int
+    created_at: int
+    updated_at: int
+    created_by: str
+    text: str
+    parent_comment_id: Optional[int]
+    # this should be CommentDict but i dont think it can self-ref
+    child_comments: Optional[List[Comment]]
+    reactions: List[CommentReaction]
--- a/querybook/server/const/comment.py
+++ b/querybook/server/const/comment.py
@@ -1,6 +1,5 @@
 from typing import List, Optional, TypedDict
-
-from querybook.server.models.comment import Comment, CommentReaction
+from models.comment import CommentReaction
 
 
 class CommentDict(TypedDict):
@@ -10,6 +9,5 @@ class CommentDict(TypedDict):
     created_by: str
     text: str
     parent_comment_id: Optional[int]
-    # this should be CommentDict but i dont think it can self-ref
-    child_comments: Optional[List[Comment]]
+    child_comment_count: int
     reactions: List[CommentReaction]
--- a/querybook/server/datasources/comment.py
+++ b/querybook/server/datasources/comment.py
@@ -3,9 +3,9 @@ from app.datasource import register
 from logic import comment as logic
 from querybook.server.app.auth.permission import verify_data_table_permission
 from querybook.server.app.db import DBSession
-from querybook.server.logic.comment_permission import assert_can_edit_and_delete
-from querybook.server.logic.datadoc_permission import (
-    assert_can_read as assert_can_read_datadoc,
+from querybook.server.logic.comment_permission import (
+    assert_can_edit_and_delete,
+    assert_can_read_datadoc,
 )
 
 
@@ -29,9 +29,9 @@ def get_comments_by_table_id(data_table_id: int):
     "/data_cell/<int:data_cell_id>/comment/",
     methods=["POST"],
 )
-def add_comment_to_cell(data_cell_id: int, datadoc_id: int, text):
+def add_comment_to_cell(data_cell_id: int, int, text):
     with DBSession() as session:
-        assert_can_read_datadoc(datadoc_id=datadoc_id, session=session)
+        assert_can_read_datadoc(data_cell_id=data_cell_id, session=session)
         return logic.add_comment_to_data_cell(
             data_cell_id=data_cell_id, uid=current_user.id, text=text
         )
--- a/querybook/server/datasources/comment.py
+++ b/querybook/server/datasources/comment.py
@@ -53,9 +53,7 @@ def add_comment_to_table(data_table_id: int, text):
     methods=["POST"],
 )
 def get_thread_comment(parent_comment_id: int, text):
-    return logic.get_thread_comment(
-        parent_comment_id=parent_comment_id, uid=current_user.id, text=text
-    )
+    return logic.get_thread_comment(parent_comment_id=parent_comment_id, text=text)
 
 
 @register(
--- a/querybook/server/datasources/comment.py
+++ b/querybook/server/datasources/comment.py
@@ -44,7 +44,7 @@ def add_comment_to_table(data_table_id: int, text):
 
 
 @register(
-    "/comment/<int:parent_comment_id>/",
+    "/comment/thread/<int:parent_comment_id>/",
     methods=["POST"],
 )
 def add_thread_comment(parent_comment_id: int, text):
@@ -75,7 +75,7 @@ def remove_comment(comment_id: int):
 
 # reactions
 @register(
-    "/comment/<int:comment_id>/",
+    "/comment/<int:comment_id>/reaction/",
     methods=["POST"],
 )
 def add_reaction(comment_id: int, reaction):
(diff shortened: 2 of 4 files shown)


czgu:
should i add permission checks for adding comments?

Yes, I think the user needs to have read access to the DataDoc they are commenting on, and metastore access for the table they are commenting on.

should users be allowed to delete comments? should they be completely deleted or archived (and say deleted comment or something like that in UI)

Yeah, let's support soft delete?

czgu:
did i set up the model correctly so that if a Comment was deleted, its corresponding DataTableComment or DataCellComment would be deleted? i think i did but just wanna confirm.

yeah, if it is cascade delete we should be good, you might wanna double check if deleting a comment would also delete all its threaded comment too with cascade

querybook/server/logic/comment.py
```diff
@@ -0,0 +1,142 @@
...
+@with_session
+def get_comment_by_id(comment_id: int, session=None):
+    return Comment.get(id=comment_id, session=session)
+
+
+@with_session
+def get_all_comment_data_dict_by_id(comment_id: int, session=None):
+    comment = Comment.get(id=comment_id, session=session)
+    child_comments = (
+        session.query(Comment).filter(Comment.parent_comment_id == comment_id).all()
```
jczhong84: can a comment have multiple child comments? If multiple, how will it be displayed on the UI?

is the purpose of this function to get only the direct children  or include all descendants? 

If only the children, will the frontend make a separate api request for each comment, in which case it seems doens't need to return the child_comments here.

If including all descendants, `filter(Comment.parent_comment_id == comment_id).all()` will not be able to do that, it can only get the children.

querybook/server/logic/comment.py
```diff
@@ -0,0 +1,142 @@
...
+            table_comment.comment_id, session=session
+        )
+        comments.append(comment_dict)
+
+    return comments
+
+
+@with_session
+def add_comment_to_data_cell(
+    data_cell_id: int, uid: int, text, commit=True, session=None
```
jczhong84, line 73: when will it be called with `commit=False` and how will it be handled?

querybook/server/logic/comment_permission.py
```diff
@@ -0,0 +1,28 @@
...
+        comment = logic.get_comment_by_id(comment_id=comment_id, session=session)
+        if comment is None:
+            raise CommentDoesNotExist
+        api_assert(
+            comment.uid == current_user.id,
+            "NOT_COMMENT_AUTHOR",
+            UNAUTHORIZED_STATUS_CODE,
+        )
+    except CommentDoesNotExist:
+        api_assert(False, "COMMENT_DNE", RESOURCE_NOT_FOUND_STATUS_CODE)
```
jczhong84, line 28: what is `COMMENT_DNE`?

querybook/server/const/comment.py
```diff
@@ -0,0 +1,15 @@
...
+
+class CommentDict(TypedDict):
+    id: int
+    created_at: int
+    updated_at: int
+    created_by: str
+    text: str
+    parent_comment_id: Optional[int]
+    # this should be CommentDict but i dont think it can self-ref
+    child_comments: Optional[List[Comment]]
```
jczhong84, line 14: seems it will work by adding `from __future__ import annotations`

querybook/server/logic/comment.py
```diff
@@ -0,0 +1,147 @@
...
+    )
+    reactions = (
+        session.query(CommentReaction)
+        .filter(CommentReaction.comment_id == comment_id)
+        .all()
+    )
+    comment_dict = comment.to_dict()
+    if child_comments:
+        comment_dict["child_comments"] = [
+            child_comment.to_dict() for child_comment in child_comments
```
jczhong84, line 28: does it also need to populate the reactions for the child comments?

Also from the UI mockup, it seems we only need the count of the child comments first. And get the details when they click the "2 replies"

querybook/server/logic/comment.py
meowcodes: it's only single level nesting

querybook/server/logic/comment_permission.py
meowcodes: do not enter or do not engage. convention used in Querybook also see DOC_DNE and BOARD_DNE

querybook/server/logic/comment.py
meowcodes: ah that makes sense. let me update that

querybook/server/datasources/comment.py
```diff
@@ -0,0 +1,93 @@
...
+    methods=["POST"],
+)
+def add_comment_to_cell(data_cell_id: int, text):
+    return logic.add_comment_to_data_cell(
+        data_cell_id=data_cell_id, uid=current_user.id, text=text
+    )
+
+
+@register(
+    "/comment/<int:data_table_id>/",
```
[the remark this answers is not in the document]
meowcodes: hmm i think you were looking at an earlier commit - fixed in the last one!

querybook/server/datasources/comment.py
```diff
@@ -0,0 +1,93 @@
...
+        parent_comment_id=parent_comment_id, uid=current_user.id, text=text
+    )
+
+
+@register(
+    "/comment/<int:comment_id>/",
+    methods=["PUT"],
+)
+def edit_comment(comment_id: int, text):
+    with DBSession() as session:
```
jczhong84, line 61: `with DBSession() as session` is not needed for flask apis, it will always get the same db session for a request.

same below

querybook/server/datasources/comment.py
```diff
@@ -0,0 +1,93 @@
...
+    return logic.add_thread_comment(
+        parent_comment_id=parent_comment_id, uid=current_user.id, text=text
+    )
+
+
+@register(
+    "/comment/<int:comment_id>/",
+    methods=["PUT"],
+)
+def edit_comment(comment_id: int, text):
```
jczhong84, line 60: add typing for text

querybook/server/datasources/comment.py
```diff
@@ -0,0 +1,93 @@
...
+        assert_can_edit_and_delete(comment_id=comment_id, session=session)
+        return logic.remove_comment(comment_id=comment_id)
+
+
+# reactions
+@register(
+    "/comment/<int:comment_id>/reaction/",
+    methods=["POST"],
+)
+def add_reaction(comment_id: int, reaction):
```
jczhong84, line 81: add typing for `reaction`

querybook/server/logic/comment.py
```diff
@@ -0,0 +1,147 @@
...
+        ]
+    comment_dict["reactions"] = [reaction.to_dict() for reaction in reactions]
+
+    return comment_dict.sort(key=lambda element: element["created_at"])
```
jczhong84, line 32: `coment_dict` is a `dict`, I dont think it has `sort` method.

if you're trying to sort the child comments, it's already ordered by created_at

querybook/server/logic/comment.py
```diff
@@ -0,0 +1,147 @@
...
+    )
+
+    comments = []
+    for cell_comment in cell_comments:
+        comment_dict = get_all_comment_data_dict_by_id(
+            cell_comment.comment_id, session=session
+        )
+        comments.append(comment_dict)
+
+    return comments
```
jczhong84, line 50: use list comprehension to simplify 
comments = [get_all_comment_data_dict_by_id(cell_comment.comment_id, session=session) for cell_comment in cell_comments]

querybook/server/logic/comment.py
```diff
@@ -0,0 +1,147 @@
...
+        .filter(DataTableComment.data_table_id == data_table_id)
+        .all()
+    )
+
+    comments = []
+    for table_comment in table_comments:
+        comment_dict = get_all_comment_data_dict_by_id(
+            table_comment.comment_id, session=session
+        )
+        comments.append(comment_dict)
```
jczhong84, line 66: same here: use list comprehension to simplify

querybook/server/datasources/comment.py
```diff
@@ -0,0 +1,93 @@
+from flask_login import current_user
+from app.datasource import register
+from logic import comment as logic
+from querybook.server.app.db import DBSession
+from querybook.server.logic.comment_permission import assert_can_edit_and_delete
+
+
+@register(
+    "/comment/data_cell/<int:data_cell_id>/",
```
czgu, line 9: the order for rest API should follow
```
/data_cell/<int:data_cell_id>/comment/
```
same as the endpoints below

PR #1264, opened 2023-06-08
pass user_id to the ai help functions

jczhong84:
the actual ai helper function may be executed in another thread because of streaming, so it wont be able to get the current_user from flask_login. Here we're passing it over.

### DIFF
--- a/querybook/server/datasources/ai_assistant.py
+++ b/querybook/server/datasources/ai_assistant.py
@@ -8,6 +8,6 @@ from lib.ai_assistant import ai_assistant
 @register("/ai/query_title/", methods=["POST"], custom_response=True)
 def generate_query_title(query):
     title_stream = ai_assistant.generate_title_from_query(
-        query=query, user_id=current_user.id if current_user else None
+        query=query, user_id=current_user.id
     )
     return Response(title_stream, mimetype="text/event-stream")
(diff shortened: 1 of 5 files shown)


querybook/server/lib/ai_assistant/ai_assistant.py
```diff
@@ -10,6 +11,10 @@ def __init__(self, provider: str, config: dict = {}):
         self._assisant = get_ai_assistant_class(provider)
         self._assisant.set_config(config)
 
+    @property
+    def current_user_id(self):
```
czgu, line 15: this ties AIAssistant to only work within Flask API, what if we want to invoke it not in web server? Can you make generate_title_from_query take in user id?

querybook/server/datasources/ai_assistant.py
```diff
@@ -1,10 +1,13 @@
...
 from app.datasource import register
 
 from lib.ai_assistant import ai_assistant
 
 
 @register("/ai/query_title/", methods=["POST"], custom_response=True)
 def generate_query_title(query):
-    title_stream = ai_assistant.generate_title_from_query(query)
+    title_stream = ai_assistant.generate_title_from_query(
+        query=query, user_id=current_user.id if current_user else None
```
czgu, line 11: this endpoint is auth'ed, you will always have current_user defined

requirements/ai/langchain.txt
```diff
@@ -1 +1 @@
-langchain==0.0.178
+langchain[openai]==0.0.178
```
czgu, line 1: good time to upgrade this 0.0.194?

czgu (approved):

PR #1265, opened 2023-06-10
feat: add error handling and use EventSource for AI assistant

jczhong84:
1. send errors through the event stream and show it on frontend with toast
2. adopted the SSE stream format
3. using EventSource on frontend to better handling the stream, instead of fetch

### DIFF
--- a/querybook/server/datasources/ai_assistant.py
+++ b/querybook/server/datasources/ai_assistant.py
@@ -5,9 +5,10 @@ from app.datasource import register
 from lib.ai_assistant import ai_assistant
 
 
-@register("/ai/query_title/", methods=["POST"], custom_response=True)
+@register("/ai/query_title/", custom_response=True)
 def generate_query_title(query):
     title_stream = ai_assistant.generate_title_from_query(
         query=query, user_id=current_user.id
     )
+
     return Response(title_stream, mimetype="text/event-stream")
--- a/querybook/server/lib/ai_assistant/base_ai_assistant.py
+++ b/querybook/server/lib/ai_assistant/base_ai_assistant.py
@@ -74,7 +74,7 @@ class BaseAIAssistant(ABC):
                 if callback_handler:
                     callback_handler.stream.send_error(err_msg)
                 else:
-                    raise Exception(err_msg)
+                    raise Exception(err_msg) from e
 
         return wrapper
 
--- a/querybook/server/lib/ai_assistant/base_ai_assistant.py
+++ b/querybook/server/lib/ai_assistant/base_ai_assistant.py
@@ -69,18 +69,21 @@ class BaseAIAssistant(ABC):
                 return func(self, *args, **kwargs)
             except Exception as e:
                 LOG.error(e, exc_info=True)
+                err_msg = self._get_error_msg(e)
                 callback_handler = kwargs.get("callback_handler")
-                err_msg = self._get_error_msg(e) or str(e.args[0])
-                callback_handler.stream.send_error(err_msg)
-                return None
+                if callback_handler:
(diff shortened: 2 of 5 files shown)


querybook/server/datasources/ai_assistant.py
```diff
@@ -5,9 +5,10 @@
 from lib.ai_assistant import ai_assistant
 
 
-@register("/ai/query_title/", methods=["POST"], custom_response=True)
+@register("/ai/query_title/", custom_response=True)
```
czgu, line 8: is it possible to keep this as post?

querybook/server/lib/ai_assistant/base_ai_assistant.py
```diff
@@ -1,12 +1,18 @@
...
+from pydantic.error_wrappers import ValidationError
+
+from lib.logger import get_logger
 
+LOG = get_logger(__file__)
 
-class ThreadedGenerator:
-    """Generator to facilitate streaming result from Langchain."""
+
+class EventStream:
```
czgu, line 13: is this event stream generic? if so, can you put this into lib or datasource/?

querybook/server/lib/ai_assistant/base_ai_assistant.py
```diff
@@ -50,17 +62,37 @@ def name(self) -> str:
...
         self._config = config
 
+    def catch_error(func):
+        @functools.wraps(func)
+        def wrapper(self, *args, **kwargs):
+            try:
+                return func(self, *args, **kwargs)
+            except Exception as e:
+                LOG.error(e, exc_info=True)
+                callback_handler = kwargs.get("callback_handler")
```
czgu, line 72: callback_handler is required, can you make it more explicit (not just in except)
it would be useful to note that callback_handler must be passed as kwargs not args

querybook/server/lib/ai_assistant/base_ai_assistant.py
```diff
@@ -50,17 +62,37 @@ def name(self) -> str:
...
+                err_msg = self._get_error_msg(e) or str(e.args[0])
```
czgu, line 73: can you make _get_error_msg just handle the or case as well?

querybook/server/lib/ai_assistant/base_ai_assistant.py
```diff
@@ -50,17 +62,37 @@ def name(self) -> str:
     def set_config(self, config: dict):
         self._config = config
 
+    def catch_error(func):
```
czgu, line 65: catch_error_to_stream?

querybook/server/datasources/ai_assistant.py
jczhong84: SSE doesn't support post method. if we'd like to keep post, we can switch back to use `fetch` and parse it accordingly.

querybook/server/lib/ai_assistant/base_ai_assistant.py
jczhong84: It'll be only needed by the ai assistant for now. I'd keep it here and move them when other use cases come up in the future.

querybook/server/lib/ai_assistant/base_ai_assistant.py
jczhong84: actually callback_handler is not always required, only when it's a streaming function. and catch_error will also handle non streaming functions as well. will update

querybook/server/lib/ai_assistant/base_ai_assistant.py
jczhong84: updated the catch_error to also handle non streaming use cases.

czgu (approved):

PR #1276, opened 2023-06-22
feat: add LLM powered text2sql support

jczhong84:
- For new query generation, it will ask for selecting tables first.
 - For query editing, it will parse the tables from the query.

https://github.com/pinterest/querybook/assets/8308723/3bd096bf-6235-42ac-8420-0063d4632413

### DIFF
--- a/querybook/server/datasources/ai_assistant.py
+++ b/querybook/server/datasources/ai_assistant.py
@@ -3,6 +3,7 @@ from flask_login import current_user
 from app.datasource import register
 
 from lib.ai_assistant import ai_assistant
+from logic import datadoc as datadoc_logic
 
 
 @register("/ai/query_title/", custom_response=True)
@@ -28,11 +29,13 @@ def query_auto_fix(query_execution_id):
 def generate_sql_query(
     query_engine_id: int, tables: list[str], question: str, data_cell_id: int = None
 ):
+    data_cell = datadoc_logic.get_data_cell_by_id(data_cell_id)
+    original_query = data_cell.context if data_cell else None
     res_stream = ai_assistant.generate_sql_query(
         query_engine_id=query_engine_id,
         tables=tables,
         question=question,
-        data_cell_id=data_cell_id,
+        original_query=original_query,
         user_id=current_user.id,
     )
 
--- a/querybook/server/datasources/ai_assistant.py
+++ b/querybook/server/datasources/ai_assistant.py
@@ -22,3 +22,18 @@ def query_auto_fix(query_execution_id):
     )
 
     return Response(res_stream, mimetype="text/event-stream")
+
+
+@register("/ai/generate_query/", custom_response=True)
+def generate_sql_query(
+    query_engine_id: int, tables: list[str], question: str, data_cell_id: int = None
+):
+    res_stream = ai_assistant.generate_sql_query(
+        query_engine_id=query_engine_id,
+        tables=tables,
+        question=question,
+        data_cell_id=data_cell_id,
+        user_id=current_user.id,
+    )
+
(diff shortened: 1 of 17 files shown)


querybook/server/datasources/ai_assistant.py
```diff
@@ -22,3 +22,18 @@ def query_auto_fix(query_execution_id):
     )
 
     return Response(res_stream, mimetype="text/event-stream")
+
+
+@register("/ai/generate_query/", custom_response=True)
+def generate_sql_query(
+    query_engine_id: int, tables: list[str], question: str, data_cell_id: int = None
```
czgu, line 29: instead of data cell, use original_query so it is not limited to data doc?

querybook/webapp/components/AIAssistant/AIModeSelector.tsx
```diff
@@ -0,0 +1,41 @@
...
+    EDIT = 'EDIT',
+}
+
+interface IAIModeSelectorProps {
+    aiMode: AIMode;
+    aiModes: AIMode[];
+    onModeSelect: (mode: AIMode) => any;
+}
+
+export const AIModeSelector: React.FC<IAIModeSelectorProps> = ({
```
czgu, line 17: this is not AIMode, more like TextToSQLMode or SQLGenerationMode?

querybook/webapp/components/AIAssistant/QueryGenerationModal.tsx
```diff
@@ -0,0 +1,233 @@
...
+        queryEngineById[engineId]?.language
+    );
+    const [question, setQuestion] = useState<string>('');
+    const [tables, setTables] = useState(tablesInQuery);
+    const [aiMode, setAIMode] = useState(
+        !!query ? AIMode.EDIT : AIMode.GENERATE
+    );
+
+    useEffect(() => {
+        setTables([...new Set([...tablesInQuery, ...tables])]);
```
czgu, line 73: use https://lodash.com/docs/3.10.1#uniq with tables.concat(tablesInQuery), so the order isn't shuffled

querybook/webapp/components/AIAssistant/QueryGenerationModal.tsx
```diff
@@ -0,0 +1,233 @@
...
+            query_engine_id: engineId,
+            tables: tables,
+            question: question,
+            data_cell_id: aiMode === AIMode.EDIT ? dataCellId : undefined,
+        }
+    );
+
+    const { explanation, query: newQuery } = streamData;
+
+    const onKeyDown = (event: React.KeyboardEvent) => {
```
czgu, line 88: nit: useCallback

querybook/webapp/components/AIAssistant/QueryGenerationModal.tsx
```diff
@@ -0,0 +1,233 @@
...
+                    type: 'text',
+                    onKeyDown,
+                    ref: inputRef,
+                    autoFocus: true,
+                }}
+            />
+        </div>
+    );
+
+    const bottomDOM = newQuery && (
```
czgu, line 138: newQuery && streamStatus === StreamStatus.FINISHED?

querybook/server/datasources/ai_assistant.py
jczhong84: The problem is I realized that for GET http request, it has length limit on the query string. So it may not be safe to pass along the whole query.

Can reconsider the EventSource solution in future and change it back to use POST method. For this specific case, I can move the get query from the data cell into this api endpoint and make `ai_assistant` to accept the original query.

czgu (approved):

PR #1279, opened 2023-06-23
feat: data cell + table comments

meowcodes:
Support for  comment and comment threads
Each comment can be edited after posted
Each comment can be attached to a datadoc cell, or a whole table 
When a cell is cloned or copied, the comment is not copied. When a cell is cut, the comment is moved
Support for reactions, like Slack
Support for richtext in comments

https://github.com/pinterest/querybook/assets/29313935/dfe2f7c4-7912-49f8-8214-a54b227beec3

Screenshot 2023-06-21 at 5 17 34 PM

AP-3433


querybook/server/models/datadoc.py
```diff
@@ -117,6 +117,12 @@ class DataCell(Base):
         order_by="QueryExecution.id.desc()",
     )
 
+    comments = relationship(
+        "Comment",
+        secondary="data_cell_comment",
+        backref=backref("data_cell", uselist=False),
```
jczhong84, line 123: should it be `uselist=True` as a datadoc can have multiple comments?

also I dont see this `comments` or `data_cell` is used anywhere.

querybook/server/models/metastore.py
```diff
@@ -207,6 +207,11 @@ class DataTable(CRUDMixin, TruncateString("name", "type", "location"), Base):
     ownership = relationship(
         "DataTableOwnership", uselist=False, cascade="all, delete", passive_deletes=True
     )
+    comments = relationship(
```
jczhong84, line 210: same here

jczhong84 (approved):

PR #1280, opened 2023-06-23
fix: delta stream  parser

jczhong84:
Previously when the stream ends with non-empty buffer, the data in the buffer will not be flushed to the final result.

### DIFF
--- a/querybook/webapp/lib/stream.ts
+++ b/querybook/webapp/lib/stream.ts
@@ -41,26 +41,12 @@ export class DeltaStreamParser {
         this._currentValue = '';
         this._isPartialKey = false;
     }
-
-    public parse(delta: string): { [key: string]: string } {
-        this._parse(delta);
+    public get result() {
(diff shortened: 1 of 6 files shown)


querybook/webapp/lib/stream.ts
```diff
@@ -48,6 +48,18 @@ export class DeltaStreamParser {
         return { ...this._result };
     }
 
+    public getResult(ended: boolean): { [key: string]: string } {
```
czgu, line 51: I am a bit iffy about having mutation in getter. What about just rename this to ?
```
public close() { ... }

```
then at the end you do:
```
parser.close();
onStreamingEnd?.(parser.result);
```
also change result into a getter like
```
public get result() { return {...this._result} }
```
then the parse code can be:
```
parser.parse(delta);
onStreaming(parser.result);
```

czgu (approved):

PR #1287, opened 2023-06-26
fix: only handle the query generation keydown event when not streaming

jczhong84:
add support of streaming cancellation
Capture-2023-06-26-165438
Capture-2023-06-26-172035

also only handle the query generation keydown event when not streaming

### DIFF
--- a/querybook/webapp/components/AIAssistant/AutoFixButton.tsx
+++ b/querybook/webapp/components/AIAssistant/AutoFixButton.tsx
@@ -9,6 +9,8 @@ import { Message } from 'ui/Message/Message';
 import { Modal } from 'ui/Modal/Modal';
 import { AccentText } from 'ui/StyledText/StyledText';
 
+import './AutoFixButton.scss';
+
 interface IProps {
     query: string;
     queryExecutionId: number;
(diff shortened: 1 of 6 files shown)


querybook/webapp/components/AIAssistant/AutoFixButton.tsx
```diff
@@ -35,37 +37,53 @@ export const AutoFixButton = ({
...
+            <div className="right-align mb16">
+                <Button
+                    title="Stop Generating"
+                    color="light"
+                    onClick={cancelStream}
+                    className="mr8"
+                />
+            </div>
+        ) : (
+            fixedQuery && (
```
czgu, line 51: replace with `streamStatus === StreamStatus.DONE &&` ?

querybook/webapp/components/AIAssistant/QueryGenerationModal.tsx
```diff
@@ -91,7 +91,10 @@ export const QueryGenerationModal = ({
 
     const onKeyDown = useCallback(
         (event: React.KeyboardEvent) => {
-            if (matchKeyPress(event, 'Enter')) {
+            if (
+                streamStatus !== StreamStatus.STREAMING &&
+                matchKeyPress(event, 'Enter')
```
czgu, line 96: can you also make the textbox not editable while streaming?

querybook/webapp/components/AIAssistant/AutoFixButton.tsx
jczhong84: it doesn't always have the fixedQuery when the streaming is done. for those cases, we dont want to show the apply/cancel buttons

czgu (approved):

PR #1290, opened 2023-06-28
fix: some small changes of ai assistant

jczhong84:
1. use text area instead of input for the text2sql question to support multiple lines
2. add another button of "apply and run" to run the query directly
3. generate a title along with the query generation and query cell title will be updated once applied
4. updated the title generation prompt. Seems it will not add quotes and final period with gpt3.5 anymore.
5. updated the text2sql prompt to include the question being asked in a comment of the query
6. pass query cell id instead of the query for title generation
7. log cell id for title generation and auto fix


querybook/webapp/components/AIAssistant/QueryGenerationModal.tsx
```diff
@@ -189,6 +184,24 @@ export const QueryGenerationModal = ({
                 }}
                 color="confirm"
             />
+            <Button
+                title="Apply and Run"
```
czgu, line 188: can you add Apply & Run for Autofix as well?

czgu (commented):
lgtm, if I recall correctly, the title change is debounced. So when user clicks on title generation, would they get the latest version of the query in the backend?

querybook/webapp/components/DataDocQueryExecutions/DataDocQueryExecutions.tsx
```diff
@@ -24,136 +24,155 @@ interface IProps {
     isQueryCollapsed: boolean;
 
     changeCellContext?: (context: string) => void;
+    onChangeFromAI?: (query: string, run: boolean) => void;
```
czgu, line 27: why don't we just use changeCellContext but also add an optional parameter to run?

querybook/webapp/components/QueryCellTitle/QueryCellTitle.tsx
```diff
@@ -45,7 +46,10 @@ export const QueryCellTitle: React.FC<IQueryCellTitleProps> = ({
         }
     }, [streamStatus, title]);
 
-    const handleTitleGenerationClick = useCallback(() => {
+    const handleTitleGenerationClick = useCallback(async () => {
```
czgu, line 49: use `forceSaveDataDoc` method instead of calling api directly

czgu (approved):

PR #1297, opened 2023-07-19
feat: Added datadoc and list access permissions for user groups

jij1949:
This PR takes the already implemented user groups and expands them so that they actually grant permissions to their members. These user groups can be assigned as editors for datadocs or lists, and can have read or write privileges. Groups can also be members of other groups, and therefore groups can also inherit permissions from other groups which they will then pass on to their members.

In addition, a user will always have the access level of its most permissive permission. For example, if a user is directly added to a datadoc with read-only permissions but is also a member of a group with edit permissions, the user will have edit permissions until the group is removed or the group permissions revoked.

Screenshot 2023-07-19 at 3 24 20 PM


querybook/server/logic/generic_permission.py
```diff
@@ -0,0 +1,64 @@
...
+    editors = recursive_q.alias()
+
+    q = select(
+        [
+            func.max(editors.c.id),
+            editors.c.uid,
+            func.max(editors.c.read),
+            func.max(editors.c.write),
+        ]
+    ).group_by(editors.c.uid)
```
jczhong84, line 58: this seems to be a smart use case of CTEs, could you add some comments to help other people to understand, so that it will be easier to maintain.

also, could you add a unit test for this function?

querybook/server/logic/generic_permission.py
```diff
@@ -0,0 +1,64 @@
...
+
+    Args:
+        doc_or_board_id: The ID of the DataDoc or Board.
+        editor_type: The type of editor (DataDocEditor or BoardEditor).
+        uid: The ID of the user to filter by (optional)
+        session: The database session to use.
+
+    Returns:
+        A list of tuples containing the editor ID, the group or user ID, and the most permissive read and write permissions.
+        Editors with inherited permissions have their ID set to None.
```
jczhong84, line 26: could you add more doc of what each member is in the tuple?

querybook/server/logic/board_permission.py
```diff
@@ -8,52 +8,60 @@
...
 class BoardDoesNotExist(Exception):
     pass
 
 
 @with_session
 def user_can_edit(board_id, uid, session=None):
-    board = session.query(Board).get(board_id)
-
-    if board is None:
-        raise BoardDoesNotExist()
```
jczhong84, line 21: why remove this?

querybook/server/logic/datadoc_permission.py
```diff
@@ -13,34 +19,53 @@ class DocDoesNotExist(Exception):
...
+        doc_or_board_id=doc_id,
+        editor_type=DataDocEditor,
+        uid=uid,
+        session=session,
+    )
+
+    if len(inherited_editors) == 1:
+        # Check if the editor's write privileges are true
+        if inherited_editors[0][3]:
+            return True
```
jczhong84, line 40: the user permission checking logic, or even the whole function for datadoc and board seems quite similar, wondering if we can have something resuable

querybook/server/logic/user.py
```diff
@@ -194,6 +194,8 @@ def create_or_update_user_group(user_group: UserGroup, commit=True, session=None
             [UserGroupMember(gid=group.id, uid=user_id) for user_id in members_to_add]
         )
 
+    update_es_users_by_id(group.id)
```
jczhong84, line 197: iirc user groups are not added to es right now, not sure if we should also update the logic here as well.
https://github.com/pinterest/querybook/blob/abd32a141d6235ec677561af48a9b556db764f96/querybook/server/logic/elasticsearch.py#L724

querybook/webapp/lib/data-doc/datadoc-permission.ts
```diff
@@ -64,14 +69,21 @@ export function getViewerInfo(
...
+    viewerIds: number[],
+    nonExplicitEditorPermissions: Record<number, IDataDocEditor>
 ): IViewerInfo {
-    const editor = uid in editorsByUserId ? editorsByUserId[uid] : null;
+    const editor =
+        uid in editorsByUserId
+            ? editorsByUserId[uid]
+            : uid in nonExplicitEditorPermissions
+            ? nonExplicitEditorPermissions[uid]
+            : null;
```
jczhong84, line 80: not a fan of nested ternary operators, which is difficult to read and understand. could you rewrite it to make it more readable?

jczhong84 (commented):
could you also update the version in package.json? Also it will be nice to add some docs for this new feature in the `docs_website`

querybook/server/logic/user.py
jij1949: The logic there is already updated to remove the `if not user.is_group` conditional. Since user groups are essentially the same as users, I think they should be updated in es with the conditional removed

querybook/server/logic/generic_permission.py
```diff
@@ -0,0 +1,64 @@
...
+    recursive_q = topq.union(bottomq)
+
+    editors = recursive_q.alias()
+
+    q = select(
+        [
+            func.max(editors.c.id),
+            editors.c.uid,
+            func.max(editors.c.read),
+            func.max(editors.c.write),
```
anacltd, line 56: We tested this with Postgres and it failed (as you cannot use the `max` function with a boolean). We eluded this by casting the field to `Integer`, like so:
```suggestion
            func.max(editors.c.read.Integer),
            func.max(editors.c.write.Integer),
```

querybook/server/logic/generic_permission.py
jij1949: I was unable to write unit tests for this function because I couldn't find a way to effective mock the DB since these queries need a DB to be queried against. However, we deployed these changes to our prod and test environments to test them.

querybook/server/logic/generic_permission.py
jij1949: Hi @anacltd, I'm getting an error doing it that way but I was able to work around it by adding 
`func.max(editors.c.read.cast(Integer))` instead of `func.max(editors.c.read.Integer)`. Does this way work for you?

anacltd: so I also get an error when working locally with a MySQL DB, but in production we use a Postgres DB (with which is seems to work just fine). I think the `cast` function is a good way to make it work on both, so kudos for finding this!

anacltd (approved):
awesome work!

jczhong84 (commented):
@jij1949 thanks for updating the PR. It looks good to me overall. Put some more comments. 

Could you also update the version in package.json as well? thanks!

querybook/webapp/components/BoardViewersList/BoardViewersList.tsx
```diff
@@ -80,7 +80,10 @@ export const BoardViewersList: React.FunctionComponent<IProps> = ({
 
     const handleUserSelect = React.useCallback(
         (uid: number) => {
-            if (uid in editorsByUid || uid === board.owner_uid) {
+            if (
+                uid === board.owner_uid ||
+                (uid in editorsByUid && editorsByUid[uid].id != null)
```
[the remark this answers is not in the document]
jij1949: I believe it is possible when a user's permissions have been inherited. The purpose of this check is to make sure you can still explicitly add a user to a doc or board even if they already have inherited permissions.

querybook/server/logic/board_permission.py
```diff
@@ -93,3 +84,18 @@ def assert_is_owner(board_id, session=None):
         )
     except BoardDoesNotExist:
         api_assert(False, "BOARD_DNE", RESOURCE_NOT_FOUND_STATUS_CODE)
+
+
+@with_session
+def assert_is_not_group(id, session=None):
+    editor = session.query(BoardEditor).filter_by(id=id).first()
+    if editor is None:
+        api_assert(False, "EDITOR_DNE", RESOURCE_NOT_FOUND_STATUS_CODE)
```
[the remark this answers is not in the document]
jij1949: We are given the editor id, so in order to retrieve the uid we need to first get the editor Object.

jczhong84:
lgtm. @jij1949 thanks for the updates! could you resolve the conflicts?

@czgu do you have any other comments or concern of this feature?

jij1949:
Closing PR because updated version has been merged

PR #1300, opened 2023-07-27
Implement Presto SQLGlot Optimizers

kgopal492:
* Create a `PrestoOptimizingValidator` that extends the `PrestoExplainValidator` to run validators that provide suggestions to rewrite & optimize the query the query
* New suggestions being made to make query run more quickly:
  * Use `APPROX_DISTINCT(x)` instead of `COUNT(DISTINCT x)`
  * Combine multiple `LIKE` clauses to use `REGEXP_LIKE`
  * Use `UNION ALL` instead of `UNION`
* Suggestions are made by tokenizing query using `sqlglot` library and searching for matching patterns
* Update query/validation endpoint's return type `QueryValidationResult` to allow for a suggestion type (specifies start coordinate & end coordinate that is changed, and the suggestion text)

* Updated logic in https://github.com/pinterest/querybook/pull/1298 to return start/end coordinates of text to change, and a suggestion string


querybook/server/lib/query_analysis/validation/validators/presto_optimizing_validator.py
```diff
@@ -0,0 +1,260 @@
...
+UNION_ALL_SUGGESTION = "UNION ALL"
+APPROX_DISTINCT_SUGGESTION = "APPROX_DISTINCT("
+
+UNION_ALL_MESSAGE = "Using UNION ALL instead of UNION will execute faster"
+REGEXP_LIKE_MESSAGE = (
+    "Combining multiple LIKEs into one REGEXP_LIKE will execute faster"
+)
+APPROX_DISTINCT_MESSAGE = (
+    "Using APPROX_DISTINCT(x) instead of COUNT(DISTINCT x) will execute faster"
+)
```
jczhong84, line 25: I feel like it's unnecessary to define those constants outside, but can just place them inside each validator class.

querybook/server/lib/query_analysis/validation/validators/presto_optimizing_validator.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+def _get_query_coordinate_by_index(query: str, index: int) -> Tuple[int, int]:
+    rows = query[: index + 1].splitlines(keepends=False)
+    return len(rows) - 1, len(rows[-1]) - 1
+
+
+def _tokenize_query(query: str) -> List[Token]:
+    return Trino.Tokenizer().tokenize(query)
```
jczhong84, line 34: if these two util functions are only used by the validators here, I think we can also move them into the `BaseSQLGlotValidator` class

querybook/server/lib/query_analysis/validation/validators/presto_optimizing_validator.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+class BaseSQLGlotValidator(metaclass=ABCMeta):
```
jczhong84, line 37: will this base class only be used for the presto optimizing validators?

querybook/webapp/const/queryExecution.ts
```diff
@@ -145,9 +145,12 @@ export interface IQueryExecutionNotification {
 }
 
 export interface IQueryValidationResult {
-    line: number;
-    ch: number;
+    start_line: number;
+    start_ch: number;
+    end_line: number | null;
+    end_ch: number | null;
```
jczhong84, line 151: where are `end_line` and `end_ch` used?

kgopal492: They'll be used in the upcoming UI PR (since we'll need start and end coordinates of text that needs to be changed)

querybook/server/lib/query_analysis/validation/validators/presto_optimizing_validator.py
kgopal492: Right now it's only used for Presto, but good point. I moved the `BaseSQLGlotValidator` to a separate file so it can be used for other custom implementations

jczhong84 (approved):

PR #1301, opened 2023-07-29
feat: adopt data element in text2sql prompt

jczhong84:
- If column description is empty and the column has associated data element, use data element's description as the column description
 - Add column filter support. People can filter out columns from the table schema if they are not needed.

### DIFF
--- a/querybook/server/lib/ai_assistant/base_ai_assistant.py
+++ b/querybook/server/lib/ai_assistant/base_ai_assistant.py
@@ -94,7 +94,7 @@ class BaseAIAssistant(ABC):
 
         return str(error.args[0])
 
-    def _filter_column(self, column: DataTableColumn) -> bool:
+    def _should_skip_column(self, column: DataTableColumn) -> bool:
         """Override this method to filter out columns that are not needed."""
         return False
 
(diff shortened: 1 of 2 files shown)


querybook/server/lib/ai_assistant/base_ai_assistant.py
```diff
@@ -93,6 +94,10 @@ def _get_error_msg(self, error) -> str:
 
         return str(error.args[0])
 
+    def _filter_column(self, column: DataTableColumn) -> bool:
```
czgu, line 97: i think in filter(...), returning True means don't remove. Maybe just for clarity we can rename this to `should_skip_column`

querybook/server/lib/ai_assistant/base_ai_assistant.py
```diff
@@ -135,9 +140,17 @@ def _generate_table_schema_prompt(
...
+
                 prompt += f"- Column Name: {column.name}\n"
                 prompt += f"  Data Type: {column.type}\n"
-                prompt += f"  Description: {column.description}\n"
+                if column.description:
+                    prompt += f"  Description: {column.description}\n"
+                elif column.data_elements:
+                    # use data element's description when column's description is empty
+                    # TODO: only handling the REF data element for now. Need to handle ARRAY, MAP and etc in the future.
+                    prompt += f"  Description: {column.data_elements[0].description}\n"
```
czgu, line 153: can you also include data element name here?

czgu (approved):

PR #1302, opened 2023-07-31
Implement UI for query optimization suggestions

kgopal492:
Create a new tooltip for users to accept query optimization suggestions

https://github.com/pinterest/querybook/assets/23270317/8c6676a3-dd29-4a74-8db0-7c14ee22684f


querybook/server/lib/query_analysis/validation/validators/base_sqlglot_validator.py
```diff
@@ -39,7 +39,7 @@ def _get_query_validation_result(
         start_index: int,
         end_index: int,
         suggestion: str = None,
-        validation_result_object_type=QueryValidationResultObjectType.SUGGESTION,
+        validation_result_object_type=QueryValidationResultObjectType.LINT,
```
jczhong84, line 42: why removing the`SUGGESTION` type and using `LINT`?

querybook/webapp/components/CodeMirrorTooltip/SuggestionTooltip.tsx
```diff
@@ -0,0 +1,45 @@
...
+                    color="confirm"
+                    icon="Check"
+                />
+                <Button
+                    title="Cancel"
+                    onClick={onCancel}
+                    theme="fill"
+                    color="cancel"
+                    icon="X"
+                />
```
jczhong84, line 41: will the tooltip popup disappear when the mouse is hovering away? if so, I think a cancel button is unnecessary.

querybook/webapp/components/CodeMirrorTooltip/SuggestionTooltip.tsx
```diff
@@ -0,0 +1,45 @@
...
+const SUGGESTION_MAX_LENGTH = 20;
+
+export const SuggestionTooltip: React.FunctionComponent<IProps> = ({
+    suggestionText,
+    onAcceptSuggestion,
+    onCancel,
+}) => {
+    const truncatedSuggestion =
+        suggestionText.length > SUGGESTION_MAX_LENGTH
+            ? suggestionText.slice(0, 20) + '...'
```
jczhong84, line 19: use `SUGGESTION_MAX_LENGTH` to replace `20`

querybook/webapp/components/CodeMirrorTooltip/SuggestionTooltip.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { Button } from 'ui/Button/Button';
+
+interface IProps {
+    suggestionText: string;
+    onAcceptSuggestion: () => void;
+    onCancel: () => void;
+}
+
+const SUGGESTION_MAX_LENGTH = 20;
```
jczhong84, line 10: is 20 long enough to handle the 3 use cases?

querybook/webapp/components/QueryEditor/QueryEditor.tsx
```diff
@@ -408,6 +450,42 @@ export const QueryEditor: React.FC<
...
+                        }
+                    }
+                }, 100),
+            [getSuggestionByPosition]
+        );
+
+        const onTextHover = useCallback(
+            async (editor: CodeMirror.Editor, node, e, pos, token) => {
+                onTextHoverLongDebounce(editor, node, e, pos, token);
+                onTextHoverShortDebounce(editor, node, e, pos, token);
```
jczhong84, line 484: why we need to call two debounce functions here? can you add some comments to explain it?

querybook/webapp/components/QueryEditor/QueryEditor.tsx
```diff
@@ -408,6 +450,42 @@ export const QueryEditor: React.FC<
...
         );
 
+        const onTextHoverShortDebounce = useMemo(
+            () =>
+                debounce((editor: CodeMirror.Editor, node, e, pos, _token) => {
+                    if (markerRef.current == null) {
+                        const suggestionAnnotation =
+                            getSuggestionByPosition(pos);
+                        if (suggestionAnnotation != null) {
+                            markTextAndShowTooltip(
```
jczhong84, line 460: nit: use destruction
```
const {from, to, suggestion} = suggestionAnnotation;
```

querybook/server/lib/query_analysis/validation/validators/base_sqlglot_validator.py
kgopal492: Realized that there's no need to distinguish the two since it requires the same behavior as lint (adding the yellow/red underline to the text and the error message tooltip). If the "suggestion"property is available then we'll display the tooltip to accept the suggestion

querybook/webapp/components/CodeMirrorTooltip/SuggestionTooltip.tsx
kgopal492: It's long enough to at least show the keyword that's being changed for each of the cases. (Otherwise, I thought the string could get pretty long for the `REGEXP_LIKE` suggestion)

querybook/webapp/components/QueryEditor/QueryEditor.tsx
kgopal492: added comments!

querybook/webapp/components/QueryEditor/QueryEditor.tsx
```diff
@@ -480,7 +472,10 @@ export const QueryEditor: React.FC<
 
         const onTextHover = useCallback(
             async (editor: CodeMirror.Editor, node, e, pos, token) => {
+                // Debounce asynchronous checks with a longer delay (e.g. for requesting table metadata)
                 onTextHoverLongDebounce(editor, node, e, pos, token);
+
+                // Faster checks use a shorter delay
```
jczhong84, line 478: why the suggestion requires a faster check?

kgopal492: The UI looks pretty laggy with the 600ms check (and since the function to get the suggestion isn't async, I thought a faster check would be fine)

jczhong84 (approved):

querybook/webapp/components/CodeMirrorTooltip/SuggestionTooltip.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export const SuggestionTooltip: React.FunctionComponent<IProps> = ({
+    suggestionText,
+    onAcceptSuggestion,
+}) => {
+    const truncatedSuggestion =
+        suggestionText.length > SUGGESTION_MAX_LENGTH
+            ? suggestionText.slice(0, SUGGESTION_MAX_LENGTH) + '...'
```
czgu, line 17: SUGGESTION_MAX_LENGTH - 3 so the max length is consistent

querybook/webapp/components/CodeMirrorTooltip/SuggestionTooltip.tsx
```diff
@@ -0,0 +1,36 @@
...
+interface IProps {
+    suggestionText: string;
+    onAcceptSuggestion: () => void;
+}
+
+const SUGGESTION_MAX_LENGTH = 20;
+
+export const SuggestionTooltip: React.FunctionComponent<IProps> = ({
+    suggestionText,
+    onAcceptSuggestion,
```
czgu, line 13: have onAcceptSuggestion accept suggestionText so you can useCallback bind it in SuggestionTooltip

querybook/webapp/components/CodeMirrorTooltip/SuggestionTooltip.tsx
```diff
@@ -0,0 +1,36 @@
...
+}) => {
+    const truncatedSuggestion =
```
czgu, line 15: useMemo this

querybook/webapp/components/QueryEditor/QueryEditor.tsx
```diff
@@ -408,6 +450,37 @@ export const QueryEditor: React.FC<
             [isTokenInTable, matchFunctionWithDefinition, openTableModal]
         );
 
+        const onTextHoverShortDebounce = useMemo(
```
czgu, line 453: Nice! i think it would be useful if you add a useDebouncedFn hook to reduce the boilerplate

PR #1317, opened 2023-08-24
feat: load table warnings from metastore

jczhong84:
load table warnings from metastore if there is any.

For warnings from metastore, the created_by will be leave it as empty.

### DIFF
--- a/querybook/server/logic/metastore.py
+++ b/querybook/server/logic/metastore.py
@@ -2,7 +2,7 @@ import datetime
 
 from app.db import with_session
 from const.elasticsearch import ElasticsearchItem
-from const.metastore import DataOwner
+from const.metastore import DataOwner, DataTableWarningSeverity
 from lib.logger import get_logger
 from lib.sqlalchemy import update_model_fields
 from logic.user import get_user_by_name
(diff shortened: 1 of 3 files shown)


querybook/server/logic/metastore.py
```diff
@@ -320,6 +320,37 @@ def create_table_information(
...
+
+    For warnings from metastore, created_by will be None.
+    """
+    # delete all warnings without created_by from the table
+    session.query(DataTableWarning).filter_by(
+        DataTableWarning.created_by is None
+    ).delete()
+
+    # add warnings from metastore to the table
+    for severiy, message in warnings:
```
kgopal492, line 340: nit: severity

querybook/server/logic/metastore.py
```diff
@@ -320,6 +320,37 @@ def create_table_information(
...
+    warnings: tuple[DataTableWarningSeverity, str] = [],
+    commit=False,
+    session=None,
+):
+    """This function is used for loading table warnings from metastore.
+
+    For warnings from metastore, created_by will be None.
+    """
+    # delete all warnings without created_by from the table
+    session.query(DataTableWarning).filter_by(
```
kgopal492, line 335: should we also be filtering by the table_id here?

jczhong84: good catch!

kgopal492 (approved):

PR #1324, opened 2023-09-14
fix: added new "preformatted" transform option to properly display newlines in query results

jij1949:
This also includes a new "multiline" column type which is assigned when a string contains newlines (\n)

Screenshot 2023-09-14 at 11 03 44 AM

### DIFF
--- a/querybook/webapp/lib/query-result/transformer.tsx
+++ b/querybook/webapp/lib/query-result/transformer.tsx
@@ -81,6 +81,16 @@ const queryResultTransformers: IColumnTransformer[] = [
         transform: (v: string): React.ReactNode => v.toLocaleUpperCase(),
     },
     {
+        key: 'preformatted',
+        name: 'Preformatted',
+        appliesToType: ['multiline'],
+        priority: 0,
+        auto: true,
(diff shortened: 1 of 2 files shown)


querybook/webapp/lib/query-result/transformer.tsx
```diff
@@ -80,6 +80,16 @@ const queryResultTransformers: IColumnTransformer[] = [
...
         transform: (v: string): React.ReactNode => v.toLocaleUpperCase(),
     },
+    {
+        key: 'preformatted',
+        name: 'Preformatted',
+        appliesToType: ['multiline'],
+        priority: 0,
+        auto: true,
+        transform: (v: string): React.ReactNode => (
+            <div className={'preformatted'}>{v}</div>
```
czgu, line 90: is there associated css with preformatted?

also can you include a screenshot of what is before vs after?

jij1949: I updated the PR with this information. Thank you!

jczhong84:
lgtm. @czgu do you have any other concerns?

PR #1329, opened 2023-09-20
fix: store the task type value rather than enum when creating a task

adamstruck:
The task type was being stored in the db as `"ScheduleTaskType.PROD"` rather than the expected `"prod"`. This was causing newly created tasks to be filtered out in the `AdminTask` component ([link](https://github.com/pinterest/querybook/blob/master/querybook/webapp/components/AppAdmin/AdminTask.tsx#L57-L72)).

### DIFF
--- a/querybook/server/logic/schedule.py
+++ b/querybook/server/logic/schedule.py
@@ -65,7 +65,7 @@ def create_task_schedule(
         kwargs=kwargs or {},
         options=options or {},
         enabled=enabled,
-        task_type=get_schedule_task_type(name),
+        task_type=get_schedule_task_type(name).value,
     )
 
     session.add(schedule)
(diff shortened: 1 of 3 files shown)


querybook/server/logic/schedule.py
```diff
@@ -65,7 +65,7 @@ def create_task_schedule(
         kwargs=kwargs or {},
         options=options or {},
         enabled=enabled,
-        task_type=get_schedule_task_type(name),
+        task_type=get_schedule_task_type(name).value,
```
jczhong84, line 68: @adamstruck thanks for fixing this! could you also update here https://github.com/pinterest/querybook/blob/21c3fd4aa75ef7437792fa735fbdce3641620333/querybook/server/logic/demo.py#L75

adamstruck:
@jczhong84 anything else needed here?

jczhong84 (approved):

PR #1330, opened 2023-09-20
Add column name suggestions to presto validator

kgopal492:
Add column name suggestions to presto optimizing validator:
  1. determine if error message returned by presto explain validator is a column name error by searching the regex 
  2. if so, get invalid column name, and use elasticsearch fuzzy search to search for the correct column name in all of the tables used in the query
  3. if only 1 result is returned by the elasticsearch query (meaning that there is one possible correct column name) - return that column name as a suggestion

### DIFF
--- a/querybook/server/lib/elasticsearch/search_table.py
+++ b/querybook/server/lib/elasticsearch/search_table.py
@@ -1,8 +1,13 @@
+from typing import Dict, List, Tuple
 from lib.elasticsearch.query_utils import (
     match_filters,
     highlight_fields,
     order_by_fields,
     combine_keyword_and_filter_query,
+)
+from lib.elasticsearch.search_utils import (
+    ES_CONFIG,
+    get_matching_objects,
 )
 
 FILTERS_TO_AND = ["tags", "data_elements"]
@@ -173,3 +178,40 @@ def construct_tables_query_by_table_names(
     }
 
     return query
+
+
+def get_column_name_suggestion(
+    fuzzy_column_name: str, full_table_names: List[str]
+) -> Tuple[Dict, int]:
+    """Given an invalid column name and a list of tables to search from, uses fuzzy search to search
+    the correctly-spelled column name"""
+    should_clause = []
+    for full_table_name in full_table_names:
+        schema_name, table_name = full_table_name.split(".")
+        should_clause.append(
+            {
+                "bool": {
+                    "must": [
+                        {"match": {"name": table_name}},
+                        {"match": {"schema": schema_name}},
+                    ]
+                }
+            }
+        )
+
+    search_query = {
+        "query": {
+            "bool": {
+                "must": {
+                    "match": {
+                        "columns": {"query": fuzzy_column_name, "fuzziness": "AUTO"}
+                    }
+                },
+                "should": should_clause,
+                "minimum_should_match": 1,
+            },
+        },
+        "highlight": {"pre_tags": [""], "post_tags": [""], "fields": {"columns": {}}},
+    }
+
+    return get_matching_objects(search_query, ES_CONFIG["tables"]["index_name"], True)
--- a/querybook/server/lib/elasticsearch/search_table.py
+++ b/querybook/server/lib/elasticsearch/search_table.py
@@ -215,3 +215,25 @@ def get_column_name_suggestion(
     }
 
     return get_matching_objects(search_query, ES_CONFIG["tables"]["index_name"], True)
+
+
+def get_table_name_suggestion(fuzzy_full_table_name: str) -> Tuple[Dict, int]:
+    """Given an invalid table name use fuzzy search to search the correctly-spelled table name"""
+    schema_name, fuzzy_table_name = fuzzy_full_table_name.split(".")
+
+    search_query = {
+        "query": {
+            "bool": {
+                "must": [
+                    {"match": {"schema": schema_name}},
+                    {
+                        "match": {
+                            "name": {"query": fuzzy_table_name, "fuzziness": "AUTO"},
+                        }
+                    },
+                ]
+            }
+        },
+    }
+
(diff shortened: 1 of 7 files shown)


querybook/server/lib/query_analysis/validation/validators/presto_optimizing_validator.py
```diff
@@ -171,53 +214,117 @@ def get_query_validation_results(
...
             token_idx += 1
 
-        return validation_errors
+        return validation_results
 
 
-class PrestoOptimizingValidator(BaseQueryValidator):
-    def languages(self):
-        return ["presto", "trino"]
+class ColumnNameSuggester(BasePrestoSQLGlotDecorator):
```
jczhong84, line 220: I guess column name suggester is not specific to presto, other engines, liek sparksql can also benefit from it?

querybook/server/lib/query_analysis/validation/validators/presto_optimizing_validator.py
```diff
@@ -171,53 +214,117 @@ def get_query_validation_results(
...
-                    query, raw_tokens=query_raw_tokens
-                )
-            )
+    def _is_column_name_error(self, validation_result: QueryValidationResult) -> bool:
+        return bool(
+            re.search(r"Column .* cannot be resolved", validation_result.message)
+        )
 
-        return validation_suggestions
+    def _get_column_name_from_position(
```
jczhong84, line 239: why not just get the column name from the error message "Column .* cannot be resolved"?

querybook/server/lib/query_analysis/validation/validators/presto_optimizing_validator.py
```diff
@@ -171,53 +214,117 @@ def get_query_validation_results(
...
+            if token.start == column_error_start_index:
+                return token.text
+        return None
 
-    def _get_presto_explain_validation_results(
-        self, query: str, uid: int, engine_id: int
-    ) -> List[QueryValidationResult]:
-        return self._get_explain_validator().validate(query, uid, engine_id)
+    def _search_columns_for_suggestion(self, columns: List[str], suggestion: str):
+        """Return the case-sensitive column name by searching the table's columns for the suggestion text"""
```
jczhong84, line 253: isn't the highlighted column name always one of the columns? wondering if this function is necessary.

querybook/server/lib/query_analysis/validation/validators/presto_optimizing_validator.py
```diff
@@ -171,53 +214,117 @@ def get_query_validation_results(
...
+
+class PrestoOptimizingValidator(BaseQueryValidator):
+    def languages(self):
+        return ["presto", "trino"]
+
+    def _get_explain_validator(self):
+        return PrestoExplainValidator("")
+
+    def _get_decorated_validator(self) -> BaseQueryValidator:
+        return UnionAllValidator(
```
jczhong84, line 320: what's the reason of changing from a list of validators to a chain of validators?

querybook/server/lib/query_analysis/validation/validators/presto_optimizing_validator.py
kgopal492: The highlighted column is always lowercase, this makes sure we get the case-sensitive version of the column name

querybook/server/lib/query_analysis/validation/validators/presto_optimizing_validator.py
kgopal492: Updated to use that regex

querybook/server/lib/query_analysis/validation/validators/presto_optimizing_validator.py
kgopal492: Using the decorator pattern so that we can add suggestions on top of the validation messages

querybook/server/lib/query_analysis/validation/validators/presto_optimizing_validator.py
kgopal492: Refactored so that we can use it with other query engines

querybook/server/lib/query_analysis/validation/validators/metadata_suggesters.py
```diff
@@ -0,0 +1,150 @@
...
+    @abstractmethod
+    def table_name_error_regex(self):
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get_full_table_name_from_error(self, validation_result: QueryValidationResult):
+        raise NotImplementedError()
+
+    def _is_table_name_error(self, validation_result: QueryValidationResult) -> bool:
+        return bool(re.match(self.table_name_error_regex, validation_result.message))
```
jczhong84, line 116: these two seems to be a bit duplicate as they are running the same regex twice

how about just using `get_full_table_name_from_error` to indicate if it is a table name error? If it is not, return None, otherwise, return the table name.

or 

as it actually will do nothing when fuzzy_table_name is None inside `_suggest_table_name`, we can just rename this as `_suggest_table_name_if_needed` and remove `_is_table_name_error`

same for the column name suggestor

querybook/server/lib/elasticsearch/search_table.py
```diff
@@ -215,3 +215,25 @@ def get_column_name_suggestion(
     }
 
     return get_matching_objects(search_query, ES_CONFIG["tables"]["index_name"], True)
+
+
+def get_table_name_suggestion(fuzzy_full_table_name: str) -> Tuple[Dict, int]:
+    """Given an invalid table name use fuzzy search to search the correctly-spelled table name"""
+    schema_name, fuzzy_table_name = fuzzy_full_table_name.split(".")
```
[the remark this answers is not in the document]
kgopal492: handled the case where `.` isn't present in the provided table name.

jczhong84 (approved):
thanks for adding test cases!

PR #1339, opened 2023-09-26
feat: ask users to confirm tables before query generation

jczhong84:
When AI finds tables for the question, it generates the query using those tables directly without asking user.
Now we are changing it to ask user to confirm the tables before query generation.

https://github.com/pinterest/querybook/assets/8308723/bbaea188-9b7e-447d-91f7-17e2fbb8b5c0

### DIFF
--- a/querybook/server/lib/ai_assistant/base_ai_assistant.py
+++ b/querybook/server/lib/ai_assistant/base_ai_assistant.py
@@ -207,6 +207,8 @@ class BaseAIAssistant(ABC):
             )
             if tables:
                 socket.send_tables_for_sql_gen(tables)
+                socket.close()
+                return
 
         # not finding any relevant tables
         if not tables:
--- a/querybook/server/lib/ai_assistant/base_ai_assistant.py
+++ b/querybook/server/lib/ai_assistant/base_ai_assistant.py
@@ -199,23 +199,23 @@ class BaseAIAssistant(ABC):
         query_engine = admin_logic.get_query_engine_by_id(
             query_engine_id, session=session
         )
+
         if not tables:
-            tables = self.find_tables(
+            suggested_tables = self.find_tables(
                 metastore_id=query_engine.metastore_id,
                 question=question,
                 session=session,
             )
-            if tables:
-                socket.send_tables_for_sql_gen(tables)
-                socket.close()
-                return
 
-        # not finding any relevant tables
-        if not tables:
-            # ask user to provide table names
-            socket.send_data(
-                "Sorry, I can't find any relevant tables by the given context. Please provide table names above."
-            )
+            if suggested_tables:
+                socket.send_tables_for_sql_gen(suggested_tables)
+            else:
+                # not finding any relevant tables
+                # ask user to provide table names
+                socket.send_data(
+                    "Sorry, I can't find any relevant tables by the given context. Please provide table names above."
+                )
+
             socket.close()
             return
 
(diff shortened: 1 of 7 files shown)


querybook/server/lib/ai_assistant/base_ai_assistant.py
```diff
@@ -207,6 +207,8 @@ def generate_sql_query(
             )
             if tables:
                 socket.send_tables_for_sql_gen(tables)
+                socket.close()
```
czgu, line 210: can you rename tables to suggested_tables?

querybook/server/lib/ai_assistant/base_ai_assistant.py
```diff
@@ -207,6 +207,8 @@ def generate_sql_query(
...
+                return
```
czgu, line 211: lets just refactor the return in the next one in the same if states, so its more like if tables are not provided, no sql is getting generated

czgu: maybe a separate function would help with the readability?

querybook/webapp/components/AIAssistant/QueryGenerationModal.tsx
```diff
@@ -188,6 +197,41 @@ export const QueryGenerationModal = ({
...
+                Select the tables you would like to use or you can also search
+                for tables above.
+            </div>
+            <div className="mt8">
+                {foundTables.map((table) => (
+                    <div key={table} className="flex-row">
+                        <Checkbox
+                            value={tables.includes(table)}
+                            onChange={(checked) => {
+                                if (checked) {
```
czgu, line 213:
```ts
setTables(oldTables =>
    checked ? uniq([...]) : oldTables.filter(...)

File: querybook/webapp/components/AIAssistant/QueryGenerationModal.tsx [hunk 4]
```diff
@@ -188,6 +197,41 @@ export const QueryGenerationModal = ({
...
+                        />
+                        <TableTag
+                            metastoreId={queryEngineById[engineId].metastore_id}
+                            tableName={table}
+                        />
+                    </div>
+                ))}
+            </div>
+            <div className="mt8">
+                <Button title="Continue" onClick={onGenerate} color="confirm" />
```
czgu, line 230: Confirm & Generate SQL?

querybook/webapp/components/CodeMirrorTooltip/TableTooltip.tsx
```diff
@@ -15,13 +16,15 @@ interface IProps {
...
     schema: IDataSchema;
+    showPinItButton?: boolean;
     openTableModal?: () => any;
 }
 
 export const TableTooltip: React.FunctionComponent<IProps> = ({
     table,
     columns,
     schema,
+    showPinItButton = true,
```
czgu, line 27: just default it to false, since i saw its false below

querybook/webapp/components/CodeMirrorTooltip/TableTooltip.tsx
```diff
@@ -113,11 +116,23 @@ export const TableTooltip: React.FunctionComponent<IProps> = ({
...
+}> = ({
+    metastoreId,
+    tableFullName,
+    showPinItButton = false,
+    showDetails = true,
+}) => {
     const dispatch = useDispatch();
     const [tableId, setTableId] = useState(null);
 
+    const openTableModal = useCallback((tableId: number) => {
```
czgu, line 130: use tableId state instead of having it as a parameter

querybook/webapp/components/CodeMirrorTooltip/TableTooltip.tsx
jczhong84: this will be used by other place, like the query editor tooltip, which is true. just to not breaking them without updating them.

czgu: in this case, we can rename this to hidePinItButton and default to False, and have whatever is calling TableTooltipByName to pass that as True

czgu (approved):

PR #1347, opened 2023-10-13
enhancement: improve popover positioning logic to ensure the content is visible in the viewport

adamstruck:


querybook/webapp/ui/Popover/Popover.tsx
```diff
@@ -136,6 +136,10 @@ export const PopoverContainer = React.forwardRef<
 
             if (mainLayout === 'top') {
                 newWrapperStyle.top = box.top - contentBox.height - arrowSize;
+                // keep the popover on screen
+                if (newWrapperStyle.top < 0) {
+                    newWrapperStyle.top = box.bottom;
+                }
```
jczhong84, line 142: shouldn't this be same as `mainLayout === 'bottom'`, which is `newWrapperStyle.top = box.bottom + arrowSize;`?

querybook/webapp/ui/Popover/Popover.tsx
```diff
@@ -146,6 +150,11 @@ export const PopoverContainer = React.forwardRef<
                 } else if (subLayout === 'right') {
                     newWrapperStyle.left =
                         box.left + box.width - contentBox.width;
+                    // keep the popover on screen
+                    if (newWrapperStyle.left < 0) {
+                        newWrapperStyle.left = box.left;
+                    }
```
jczhong84, line 156: could be just `newWrapperStyle.left = 0`?

or 
`newWrapperStyle.left = Math.max(box.left + box.width - contentBox.width, 0);`

jczhong84: Also does it also need to check the right side as well for `subLayout === 'left'` above and the else below?

jczhong84 (commented):
thanks for the PR! 

I think we'll need to revisit all the cases carefully to make sure they are all handled correctly. Basically all 4 directions might be outside the screen depends on the mainLayout and subLayout.

PR #1348, opened 2023-10-13
fix: schema filtering in table search modal

adamstruck:
This component was not functioning properly. The main problem was that this component was using the environment ID as the metastore ID.


querybook/webapp/components/Search/SearchSchemaSelect.tsx
```diff
@@ -58,11 +59,12 @@ export const SearchSchemaSelect: React.FC<ISearchSchemaSelectProps> = ({
         <AsyncSelect
             styles={tableReactSelectStyle}
             placeholder={'search schema name'}
-            value={selectedSchemaItems}
             onChange={handleUpdateSearchFilter}
             loadOptions={loadOptions}
             defaultOptions={[]}
-            noOptionsMessage={() => 'No schema found.'}
+            inputValue={searchText}
+            onInputChange={(text) => setSearchText(text)}
```
jczhong84, line 66: is these two properties and `searchText` needed? seems it is working well today without them, except the metastore id issue.

adamstruck: No they are not, I'll revert.

jczhong84 (approved):

Issue #517, opened 2021-04-16
Support metastore for BigQuery


MikhailBarsukov:
There is SQLAlchemy library supporting BigQuery dialect that can be used for that: https://github.com/googleapis/python-bigquery-sqlalchemy

PR #1352, opened 2023-10-23
Bigquery sqlalchemy metastore

seuf:
This Merge request allow the use of sqlalchemy-bigquery as a metastore for bigquery engine.

Also bump bigquery requirements files and add sqlalchemy-bigquery dependency.

Allow use of empty google credentials for bigquery engine : In case of running in GKE cluster, we can use in cluster service account with workload identity to authenticate to google.

It Fix : https://github.com/pinterest/querybook/issues/517


czgu (approved):
awesome, thank you!

PR #1370, opened 2023-11-14
fix: Pin Yarn to 1.22.19 to avoid issue in newer version

baumandm:
The latest Yarn version breaks the Querybook build, see: https://github.com/yarnpkg/yarn/issues/9011


baumandm:
Fixed upstream, closing

PR #1376, opened 2023-11-28
fix: s3 table upload location

jczhong84:
- Explicitly call out that the table is managed when `use_schema_location=False` for s3 table uploader. 
 - Get the actual table location from metastore for data uploading for managed tables
 - Fix the logic of checking `is_external`. Previously `is_external` will be `False` when `use_schema_location` is True, which supposed to be True (external table), as it is specifying location.
   ```is_external = not self._exporter_config.get("use_schema_location", False)```

### DIFF
--- a/docs_website/docs/changelog/breaking_change.md
+++ b/docs_website/docs/changelog/breaking_change.md
@@ -7,12 +7,21 @@ slug: /changelog
 
 Here are the list of breaking changes that you should be aware of when updating Querybook:
 
+## v3.29.0
+
+Made below changes for `S3BaseExporter` (csv table uploader feature):
+
+-   Both `s3_path` and `use_schema_location` are optional now
+-   If none is provided, or `use_schema_location=False`, the table will be created as managed table, whose location will be determined by the query engine.
+
 ## v3.27.0
+
 Updated properties of `QueryValidationResult` object. `line` and `ch` are replaced with `start_line` and `start_ch` respectively.
 
 ## v3.22.0
 
 Updated the charset of table `data_element` to `utf8mb4`. For those mysql db's default charset is not utf8, please run below sql to update it if needed.
+
 ```sql
 ALTER TABLE data_element CONVERT TO CHARACTER SET utf8mb4
 ```
--- a/docs_website/docs/changelog/breaking_change.md
+++ b/docs_website/docs/changelog/breaking_change.md
@@ -13,6 +13,7 @@ Made below changes for `S3BaseExporter` (csv table uploader feature):
 
 -   Both `s3_path` and `use_schema_location` are optional now
 -   If none is provided, or `use_schema_location=False`, the table will be created as managed table, whose location will be determined by the query engine.
+-   Previously `use_schema_location=True` will create managed table, and now it creates external table.
 
 ## v3.27.0
 
--- a/docs_website/docs/integrations/add_table_upload.md
+++ b/docs_website/docs/integrations/add_table_upload.md
@@ -51,7 +51,7 @@ Available options:
     Checkout here for examples in SparkSQL: https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-create-table-hiveformat.html#examples
     For Trino/Presto, it would be the WITH statement: https://trino.io/docs/current/sql/create-table.html
 
-If neither s3_path nor use_schema_location is supplied, it will be treated same as `use_schema_location=False``, and it will be created as managed table.
+If neither s3_path nor use_schema_location is supplied, it will be treated same as `use_schema_location=False`, and it will be created as managed table.
 
 ### S3 Parquet exporter
 
--- a/docs_website/docs/integrations/add_table_upload.md
+++ b/docs_website/docs/integrations/add_table_upload.md
@@ -42,16 +42,16 @@ Included by default: No
 
 Available options:
 
-Either s3_path or use_schema_location must be supplied.
-
--   s3_path (str): if supplied, will use it as the root path for upload. Must be the full s3 path like s3://bucket/key, the trailing / is optional.
--   use_schema_location (boolean):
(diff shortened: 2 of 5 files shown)


docs_website/docs/integrations/add_table_upload.md
```diff
@@ -48,6 +48,7 @@ Either s3_path or use_schema_location must be supplied.
 -   use_schema_location (boolean):
     if true, the upload root path is inferred by locationUri specified by the schema/database in HMS. To use this option, the engine must be connected to a metastore that uses
     HMSMetastoreLoader (or its derived class).
+    if false, it will be created as managed table, whose location will be determined automatically by the query engine.
```
czgu, line 51: what if it is not supplied?

See below, I think we can update the documentation to make neither required and document the behavior as such:

- (If s3_path is supplied) ....
- (If use_schema_location is supplied) ...
- (If neither are supplied) ...

querybook/server/lib/metastore/loaders/hive_metastore_loader.py
```diff
@@ -101,6 +101,12 @@ def get_partitions(
     def get_schema_location(self, schema_name: str) -> str:
         return self.hmc.get_database(schema_name).locationUri
 
+    def get_table_location(self, schema_name: str, table_name: str) -> str:
```
czgu, line 104: `-> Union[None, str]`

querybook/server/lib/table_upload/exporter/s3_exporter.py
```diff
@@ -61,40 +69,39 @@ def destination_s3_path(self, session=None) -> str:
...
-        if self._exporter_config.get("use_schema_location", False):
-            query_engine = get_query_engine_by_id(self._engine_id, session=session)
-            metastore = get_metastore_loader(query_engine.metastore_id, session=session)
+        query_engine = get_query_engine_by_id(self._engine_id, session=session)
+        metastore = get_metastore_loader(query_engine.metastore_id, session=session)
 
-            if metastore is None:
-                raise Exception("Invalid metastore")
+        if metastore is None:
+            raise Exception("Invalid metastore")
```
czgu, line 88: Invalid metastore for table upload

querybook/server/lib/table_upload/exporter/s3_exporter.py
```diff
@@ -32,13 +32,21 @@
...
+        if false, it will be created as managed table, whose location will be determined automatically by the query engine.
     - table_properties (List[str]): list of table properties passed, this must be query engine specific.
         Checkout here for examples in SparkSQL: https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-create-table-hiveformat.html#examples
         For Trino/Presto, it would be the WITH statement: https://trino.io/docs/current/sql/create-table.html
 """
 
 
 class S3BaseExporter(BaseTableUploadExporter):
+    def __init__(self, exporter_config: dict = {}):
+        if ("s3_path" not in exporter_config) and (
```
czgu, line 44: actually, we don't need use_schema_location to be provided, i guess neither needs to supplied and the default
would just be using the table location

querybook/server/lib/table_upload/exporter/s3_exporter.py
```diff
@@ -61,40 +69,39 @@ def destination_s3_path(self, session=None) -> str:
...
             if not schema_location_uri:
                 raise Exception("Invalid metastore to use use_schema_location option")
 
-            return sanitize_s3_url_with_trailing_slash(schema_location_uri)
+            return sanitize_s3_url(schema_location_uri) + "/" + table_name
 
-        raise Exception("Must specify s3_path or set use_schema_location=True")
+        # Use its actual location for managed tables
+        table_location = metastore.get_table_location(schema_name, table_name)
+        if table_location:
```
czgu, line 99: just follow the same pattern as above
```
if not table_location:
    raise Exception(...)
return sanitize_s3_url(...)
```

querybook/server/lib/table_upload/exporter/s3_exporter.py
```diff
@@ -118,14 +125,14 @@ def _handle_if_table_exists(self, session=None):
     def _get_table_create_query(self, session=None) -> str:
         query_engine = get_query_engine_by_id(self._engine_id, session=session)
         schema_name, table_name = self._fq_table_name
-        is_external = not self._exporter_config.get("use_schema_location", False)
+        is_managed = self._exporter_config.get("use_schema_location") is False
```
czgu, line 128: is_managed = "s3_path" not in self._exporter_config

querybook/webapp/components/TableUploader/TableUploaderForm.tsx
```diff
@@ -77,7 +77,11 @@ export const TableUploaderForm: React.FC<ITableUploaderFormProps> = ({
                 error: 'Fail to create table',
             });
 
-            navigateWithinEnv(`/table/${tableId}`);
+            // sometimes there will be sync delay between the metastore and querybook
+            // skip the redirect if the table has not been synced over.
+            if (tableId) {
+                navigateWithinEnv(`/table/${tableId}`);
```
czgu, line 83: I think there should be a toast send if tableId isn't available yet

docs_website/docs/changelog/breaking_change.md
```diff
@@ -7,12 +7,21 @@ slug: /changelog
 
 Here are the list of breaking changes that you should be aware of when updating Querybook:
 
+## v3.29.0
+
+Made below changes for `S3BaseExporter` (csv table uploader feature):
+
+-   Both `s3_path` and `use_schema_location` are optional now
```
czgu, line 14: you need to mention if `use_schema_location=True` before 3.29.0, it would create managed table, now it creates external

czgu (dismissed):

czgu (approved):

PR #1387, opened 2023-12-18
feat: added ability to omit clean_up jobs from general db_clean task and fixed clean_up_archived_data_doc

jij1949:
Now -1 can be passed into the `run_all_db_clean_up_jobs` task for each `days_to_keep` argument value. This will omit the specific db_clean_up tasks from being run. For example, if the value for `days_to_keep_task_record` is passed in as -1, then `clean_up_task_run_record` will not be run. 

Previously, clean_up_archived_data_doc would not run properly and the logic meant that only non-archived datadocs would be deleted, which is the inverse of the intent of the function. This has been changed to only delete archived datadocs that have been updated more than `days_to_keep` ago


czgu (approved):
Thanks!

PR #1389, opened 2023-12-19
feat: update column stats ui, add sort by usage to columns

kgopal492:
Screenshot 2023-12-19 at 11 39 36 AM

* Update column stats UI to show table column stats as tags (rather than additional rows)
* Add ability to sort columns by usage
* Query all table column details from table column page to be able to sort columns by usage

### DIFF
--- a/querybook/webapp/components/DataTableViewColumn/DataTableViewColumn.tsx
+++ b/querybook/webapp/components/DataTableViewColumn/DataTableViewColumn.tsx
@@ -21,7 +21,7 @@ export interface IDataTableViewColumnProps {
     onEditColumnDescriptionRedirect?: Nullable<() => Promise<void>>;
 }
 
-type ColumnOrderBy = 'alphabetical' | 'usage';
+type ColumnOrderBy = 'Default' | 'Aa' | 'Usage';
 
 const COLUMN_STATS_USAGE_KEY = 'usage';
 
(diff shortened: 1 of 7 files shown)


querybook/webapp/components/DataTableViewColumn/DataTableViewColumn.tsx
```diff
@@ -21,20 +21,44 @@ export interface IDataTableViewColumnProps {
...
     table = null,
-    tableColumns = [],
     numberOfRows = null,
     onEditColumnDescriptionRedirect,
 }) => {
     const [filterString, setFilterString] = React.useState('');
-    const [orderColumnsBy, setOrdeColumnsBy] = React.useState(true);
-    const [orderBoardBy, setOrderBoardBy] = React.useState(false);
+    const [orderColumnsBy, setOrderColumnsBy] =
+        React.useState<ColumnOrderBy>('usage');
```
czgu, line 38: COLUMN_STATS_USAGE_KEY?

querybook/webapp/components/DataTableViewColumn/DataTableViewColumn.tsx
```diff
@@ -21,20 +21,44 @@ export interface IDataTableViewColumnProps {
     onEditColumnDescriptionRedirect?: Nullable<() => Promise<void>>;
 }
 
+type ColumnOrderBy = 'alphabetical' | 'usage';
```
czgu, line 24: include a 'default' sort that doesn't sort?

querybook/webapp/components/DataTableViewColumn/DataTableViewColumn.tsx
czgu: if there is no usage, what would happen?

kgopal492: This 'usage' string just corresponds to the column order by type (so shouldn't use the constant here). If there's no usage, they're all treated as 0 usage. But I updated it so that you can't see the sort by usage option if there are no usages available for any of the columns so it isn't confusing to users.

czgu (approved):

PR #1390, opened 2023-12-20
feat: Added support for sending custom datadog metrics from querybook

jij1949:
This PR includes the setup of `dogstatsd`, a python library used for sending metrics to Datadog as well as the creation of a new stats logger, `DatadogStatsLogger`.

In addition, it includes 5 metrics: `celery.active_workers`, `celery.active_tasks`, `task.received`, `task.failures`, and `task.successes`

The first 2 metrics are sent to datadog every 5 seconds by opening a new thread which continuously runs the `send_stats_logger_metrics` function. The other 3 metrics are triggered by event handlers so they are sent whenever a task is received by a worker, a task fails, or a task succeeds.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,6 +20,26 @@ x-querybook-volumes: &querybook-volumes
     # - file:/opt/store/
 
 services:
+    datadog:
+        image: gcr.io/datadoghq/agent:latest
+        ports:
+            - 8125:8125/udp
+        expose:
+            - 8125/udp
+        environment:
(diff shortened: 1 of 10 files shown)


czgu (commented):
Hi Julian, thanks for submitting this PR and the code lgtm, can you make datadog not a required dependency for Querybook in dev mode? Would it work if you move those code to example documentation instead?

jij1949:
Datadog is now an optional dependency

jczhong84:
@czgu can you help to look the PR again?

PR #1392, opened 2024-01-03
Auth0 Integration V2

aidenprice:
Hi again,

This is the cleaned up PR from the previous attempt, please ignore my earlier PR and consider this one instead. Sorry for the messy git history.

This PR adds Auth0 OAuth integration following the same pattern as Okta. The key names are different enough to justify a different handler.

Thank you again,
Aiden


jczhong84:
@aidenprice thanks for the PR to add Auth0. 
Looks like it's mostly same as okta except the `get_auth0_urls`. Maybe we can just inherit it from `OktaLoginManager`,  rename `get_okta_urls` as `get_oauth_urls` and override it for `Auth0LoginManager`

aidenprice:
Hi, I've tried your suggestion and it certainly cleans up a lot of repeated code.
One thing I should point out though, the `LOG = get_logger(__file__)` isn't inheritable or injectable. So when you instantiate an `Auth0LoginManager` it will represent logs as coming from the Okta file.

querybook/server/app/auth/auth0_auth.py
```diff
@@ -0,0 +1,36 @@
+import requests
+from env import get_env_config
+
+from app.auth.okta_auth import OAUTH_CALLBACK_PATH, OktaLoginManager
+
+
+class NoopAuth(requests.auth.AuthBase):
```
jczhong84, line 7: this is not used

aidenprice: Done, removed it.

jczhong84 (approved):

aidenprice:
Thank you!

PR #1395, opened 2024-01-12
feat: Improve Cron schedule support

baumandm:
Improved support for task scheduling via cron expression by integrating it directly into the RecurrenceEditor.

Aside from simplifying the code and UI, there are a number of functional benefits:

- A description of the cron is shown thanks to [cRonstrue](https://github.com/bradymholt/cronstrue)
- You can mostly build a cron in one of the other modes, then switch into cron mode to continue editing what you had selected
- If a DataDoc schedule is manually updated to a non-supported cron expression, the DataDoc schedule modal continues to work as expected (previously it crashes)
- By default, cron schedules are still not allowed on DataDocs, but enabling it is only a one-line change

Screenshot 2024-01-12 at 2 41 18 PM


querybook/webapp/lib/utils/cron.ts
```diff
@@ -136,24 +159,46 @@ export function getRecurrenceLocalTimeString(
         .format(format);
 }
 
+/**
+ * Determines whether a cron string is supported by the recurrence editor.
+ *
+ * @param cron Cron string to validate
+ * @returns true if cron string is supported by recurrence editor, false otherwise
+ */
```
jczhong84, line 167: maybe we can also use [cron-parser](https://www.npmjs.com/package/cron-parser) for validation, which is mentioned by package `cronstrue`

jczhong84 (commented):
Thanks for the PR! It really looks more easy and clean. 

One issue when I tried:
When it is in the `weekly`, `monthly` and `yearly` tab, it can't be switched to the `cron` tab.  Got below exception
  ```
  cron.ts:97 Uncaught TypeError: Cannot read properties of undefined (reading 'join')
    at recurrenceToCron (cron.ts:97:41)
```

querybook/webapp/lib/utils/cron.ts
baumandm: The backend has cron validation that prevents saving invalid crons, so I don't think we would need this here.  This function is only used to determine if the RecurrenceEditor should use `cron` mode or one of the simple modes.

baumandm:
> One issue when I tried:
> When it is in the `weekly`, `monthly` and `yearly` tab, it can't be switched to the `cron` tab. Got below exception

Previously, you had to fill out the required fields in the `weekly`/`monthly`/`yearly` tabs before you could switch to `cron`, but I agree this doesn't make a lot of sense.  Now, if there is any issue converting the recurrence editor selections to cron, it just resets back to `""`

jczhong84 (approved):

PR #1398, opened 2024-01-19
feat: Add non-streaming support to AI Assistant

baumandm:
This PR allows toggling the [streaming mode](https://api.python.langchain.com/en/latest/chat_models/langchain_openai.chat_models.base.ChatOpenAI.html#langchain_openai.chat_models.base.ChatOpenAI.streaming) of the OpenAIAssistant.  This is required when using a proxy that doesn't support streaming.

The default behavior remains `streaming: true`, but it can now be disabled via the `model_args`:

```
AI_ASSISTANT_PROVIDER: openai
AI_ASSISTANT_CONFIG:
    default:
        model_args:
            model_name: gpt-3.5-turbo
            temperature: 0
            streaming: false
```

I added the option into the config files to help document the feature, but it's completely optional and doesn't need to be specified at all.

When disabled, the LLM result is sent in single chunk when the response is complete.

### DIFF
--- a/containers/bundled_querybook_config.yaml
+++ b/containers/bundled_querybook_config.yaml
@@ -15,6 +15,7 @@ ELASTICSEARCH_HOST: http://elasticsearch:9200
 #         model_args:
 #             model_name: gpt-3.5-turbo
 #             temperature: 0
+#             streaming: true
 #         reserved_tokens: 1024
 #     table_summary:
 #         model_args:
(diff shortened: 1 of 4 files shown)


jczhong84:
@baumandm we're looking at the new version of Langchain which will have a much simpler way for the steaming, will try to cover the non streaming case when doing so

PR #1400, opened 2024-01-21
fix: modify trigger for query_authoring survey

jczhong84:
Current:
The survey triggers on any cell execution. It means you can just open a DataDoc, hit Run button and the survey will show up asking how easy it was to create the query (and no query authoring happened). 

Updated as: 
Survey asking for query_authoring experience is shown only after user created new or modified existing query and ran it.

### DIFF
--- a/querybook/webapp/components/DataDocQueryCell/DataDocQueryCell.tsx
+++ b/querybook/webapp/components/DataDocQueryCell/DataDocQueryCell.tsx
@@ -102,7 +102,7 @@ interface IState {
     query: string;
     meta: IDataQueryCellMeta;
 
-    modified: boolean;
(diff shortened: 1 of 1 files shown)


querybook/webapp/components/DataDocQueryCell/DataDocQueryCell.tsx
```diff
@@ -102,6 +102,7 @@ interface IState {
     query: string;
     meta: IDataQueryCellMeta;
 
+    modified: boolean;
```
czgu, line 105: can you track it with a timestamp instead? so that we only trigger for past 5 minutes of action

czgu (approved):

PR #1403, opened 2024-01-31
fix: Update `unique_table_ownership` constraint to allow owner per type

baumandm:
This PR addresses a limitation of the Data Table Ownership type feature, which was previously added in #1170.  That PR added a `type` column to support multiple types of owners per table.  However, the unique constraint between `data_table_id` and `uid` prevents any user from having multiple types of ownership.

This PR adds a new database migration that updates the constraint to include the `type` column.

Screenshot 2024-01-31 at 2 03 11 PM


jczhong84 (approved):

jczhong84:
Thanks @baumandm for fixing it!

PR #1407, opened 2024-02-05
chore: Update pip install command

baumandm:
Previously, the `pip install` command in the Dockerfile ended in `|| true`, so it never failed.  I think the purpose was to make `local.txt` optional, but it also covers up any errors installing packages.

This removes `|| true` and instead makes `local.txt` optional by checking to see if the file exists.


czgu (approved):
thank you!

PR #1408, opened 2024-02-09
fix: Fix create task schedule task_type

baumandm:
Currently Data Doc tasks are being created with `task_type` `"prod"` instead of `"user"`.  

This is because the `get_schedule_task_type()` function expects the Celery task name (`tasks.run_datadoc.run_datadoc`), not the task name (`run_data_doc_#`).


jczhong84 (approved):

jczhong84:
thanks for fixing it!

PR #1413, opened 2024-02-22
fix: update schedule disabler configs

rchandnaeg:


rchandnaeg:
- adds kwarg options to the disable schedule task
- corrects behavior of `skip_if_no_impression_with_export` and `skip_if_no_impression_but_non_select` to match the semantic meaning of these options. "Skip disabling the datadoc if it has no impression but has an exporter", and "Skip disabling the datadoc if it has no impression but there are non-select queries"
- support for entering null values to kwargs, and improves rendering of non-integer default values for tasks.

Previously, creating this task looked like this:
Screenshot 2024-02-22 at 2 11 23 PM

View of creating the task now with kwarg options
Screenshot 2024-02-21 at 12 45 52 PM

PR #1414, opened 2024-02-22
fix: update schedule disabler configs

rchandnaeg:
- adds kwarg options to the disable schedule task
- corrects behavior of `skip_if_no_impression_with_export` and `skip_if_no_impression_but_non_select` to match the semantic meaning of these options. "Skip disabling the datadoc if it has no impression but has an exporter", and "Skip disabling the datadoc if it has no impression but there are non-select queries"
- support for entering null values to kwargs, and improves rendering of non-integer default values for tasks.

Previously, creating this task looked like this:
Screenshot 2024-02-22 at 2 11 23 PM

View of creating the task now with kwarg options
Screenshot 2024-02-21 at 12 45 52 PM

### DIFF
--- a/querybook/server/tasks/disable_scheduled_docs.py
+++ b/querybook/server/tasks/disable_scheduled_docs.py
@@ -279,7 +279,7 @@ def disable_scheduled_docs(
     disable_if_failed_for_n_runs=5,
     disable_if_no_impression_for_n_days=30,
     skip_if_no_impression_but_non_select=True,
-    skip_if_no_impression_with_export=True
+    skip_if_no_impression_with_export=True,
 ):
     tasks_to_disable = disable_deactivated_scheduled_docs(
         dry_run=dry_run,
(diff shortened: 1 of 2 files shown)


querybook/server/tasks/disable_scheduled_docs.py
```diff
@@ -117,12 +117,12 @@ def check_task_no_impression_for_n_days(
...
     doc_id = task_dict["kwargs"]["doc_id"]
 
-    if not disable_config.skip_if_no_impression_with_export:
+    if disable_config.skip_if_no_impression_with_export:
         task_have_export = len(task_dict["kwargs"].get("exports", [])) > 0
         if task_have_export:
             return False
 
-    if not disable_config.skip_if_no_impression_but_non_select:
+    if disable_config.skip_if_no_impression_but_non_select:
```
jczhong84, line 125: @czgu could you help to look at this? I'm a bit confused about the actual meaning of these two configs. should it disable or not when they are true.
skip_if_no_impression_with_export
skip_if_no_impression_but_non_select

PR #1418, opened 2024-03-01
fix: add native sum column analyzer

rchandnaeg:
I know that there's the option to add custom column analyzers, but I think sum is common enough to be supported natively and appear at the top of the Quick Insights view of a column.

Screenshot 2024-03-01 at 11 42 57 AM


czgu (commented):
thank you! great idea

czgu (approved):

PR #1419, opened 2024-03-04
feat: Clone DataDocs into other Environments

baumandm:
Adds support for cloning a DataDoc into another environment.

- Refactored the Clone button logic into a sub component of the DataDocRightSidebar
- Added a `<Select>` with the environments available to the user
- Defaults to the current environment
- Added an optional `POST` parameter `environment_id` to the clone endpoint
- The API call is backwards-compatible, defaults to the DataDoc's environment as before
- Redirects to the DataDoc in the selected environment after cloning.

Screenshot 2024-03-04 at 1 16 59 PM

The cloned DataDoc may use Query Engines that are not available in the selected environment, which gives `Invalid Query Cell` warnings as shown below.  This is **not** automatically adjusted during the clone process, it needs to be manually fixed after cloning.  IMHO this is preferable since it ensures a human will review the selected Query Engines for correctness.

Screenshot 2024-03-04 at 1 50 11 PM

### DIFF
--- a/querybook/webapp/components/DataDocRightSidebar/DataDocCloneButtonConfirm.tsx
+++ b/querybook/webapp/components/DataDocRightSidebar/DataDocCloneButtonConfirm.tsx
@@ -1,0 +1,66 @@
+import React, { useCallback } from 'react';
+import { useSelector } from 'react-redux';
+
+import { availableEnvironmentsSelector } from 'redux/environment/selector';
+import { IEnvironment } from 'redux/environment/types';
+import { makeSelectOptions, Select } from 'ui/Select/Select';
+import { StyledText } from 'ui/StyledText/StyledText';
+
+interface IProps {
+    defaultEnvironment: IEnvironment;
+    onEnvironmentChange: (environment: IEnvironment) => void;
+}
+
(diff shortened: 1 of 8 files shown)


querybook/webapp/components/DataDocRightSidebar/DataDocCloneButtonConfirm.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { IEnvironment } from 'redux/environment/types';
+import { makeSelectOptions, Select } from 'ui/Select/Select';
+import { StyledText } from 'ui/StyledText/StyledText';
+
+interface IProps {
+    defaultEnvironment: IEnvironment;
+    onEnvironmentChange: (environment: IEnvironment) => void;
+}
+
+export const DataDocCloneButtonConfirm: React.FunctionComponent<IProps> = ({
```
jczhong84, line 14: this component name seems a bit weird to me. I think this component could be just an environment selector, and move the messages to the DataDocCloneButton component.

jczhong84 (commented):
thanks for the PR @baumandm !

covalesj (approved):

PR #1421, opened 2024-03-07
feat: add table sampling support

jczhong84:
Here we add two approaches of table sampling support:
- Sampled Table
There will be a new separated sampled version of the original table being created.
- Bucketing
There will be no new table, but the original table will be reorganized so that it can be queried with additional filters from a small portion of the dataset. This will be using the official `TABLESAMPLE` clause from the language. 

- If a table support sampling, we can add it in the column of `custom_properties` of table `data_table_information`
  - sampling: bool  
Indicating if this table supports sampling or not.
  - sampled_table: str 
 Name of the sampled version of the table. This has higher priotiry. If sampled_table is not provided while sampling is true, we'll use the `TABLESAMPLE` clause.

- If any table in the query support sampling, we'll show the `Sample` selector on the right side of the query engine selector
- If a sample rate is selected, we'll replace the table with either with sampled table or use the `TABLESAMPLE` clause to transform the query before running.

https://github.com/pinterest/querybook/assets/8308723/7f83d7cd-4186-4880-93fe-1ebbac512133


querybook/server/lib/query_analysis/transform.py
```diff
@@ -0,0 +1,169 @@
...
+        read=dialect,
+        write=dialect,
+        pretty=True,
+    )
+
+    return _statements_to_query(statements)
+
+
+def get_select_statement_limit(
+    statement: Union[exp.Expression, str],
```
jczhong84, line 35: added those functions for query limiting, but has not been used.

querybook/webapp/components/DataDocQueryCell/DataDocQueryCell.tsx
```diff
@@ -644,6 +675,20 @@ class DataDocQueryCellComponent extends React.PureComponent<IProps, IState> {
         );
     }
 
+    @bind
+    public async handleTablesChange(tables: Record<string, IDataTable>) {
```
czgu, line 679: use tablesByName or change the type annotation to indicate what the key is

querybook/webapp/components/QueryRunButton/QueryRunButton.tsx
```diff
@@ -93,6 +102,15 @@ export const QueryRunButton = React.forwardRef<
...
 
+        const tableSamplingDOM =
+            !disabled && TableSamplingConfig.enabled && hasSamplingTables ? (
+                <TableSamplingSelector
+                    sampleRate={sampleRate}
+                    setSampleRate={onSampleRateChange}
+                    tooltipPos={runButtonTooltipPos}
+                />
+            ) : null;
+
```
czgu, line 113: can this live outside of QueryRunButton? I dont think it should control the UI for TableSamplingSelector, that's the job of the datacell or querycomposer

querybook/webapp/components/QueryRunButton/QueryRunButton.tsx
```diff
@@ -249,3 +268,60 @@ const QueryLimitSelector: React.FC<{
         </Dropdown>
     );
 };
+
+const TableSamplingConfig = PublicConfig.table_sampling ?? {
```
czgu, line 272: TABLE_SAMPLING_CONFIG

just for consistent naming as below

querybook/webapp/components/QueryRunButton/QueryRunButton.tsx
```diff
@@ -249,3 +268,60 @@ const QueryLimitSelector: React.FC<{
...
+    default_sample_rate: 0,
+};
+const sampleRateOptions = [{ label: 'none', value: 0 }].concat(
+    TableSamplingConfig.sample_rates.map((value) => ({
+        label: value + '%',
+        value,
+    }))
+);
+const DEFAULT_SAMPLE_RATE = TableSamplingConfig.default_sample_rate;
+const TableSamplingSelector: React.FC<{
```
czgu, line 284: move to another component file

querybook/webapp/const/metastore.ts
```diff
@@ -73,7 +73,7 @@ export interface IDataTable {
     column_info?: {
         partition_keys?: string[];
     };
-    custom_properties?: Record<string, string>;
+    custom_properties?: Record<string, any>;
```
czgu, line 76: string | number?

requirements/parser/sqlglot.txt
```diff
@@ -1 +1 @@
-sqlglot==17.8.0
+sqlglot[rs]==22.1.1
```
czgu, line 1: remove this file?

querybook/server/lib/query_analysis/transform.py
```diff
@@ -0,0 +1,169 @@
...
+        return None
+
+    limit = -1
+    limit_clause = statement.args.get("limit")
+
+    if isinstance(limit_clause, exp.Limit):
+        limit = limit_clause.expression.this
+    elif isinstance(limit_clause, exp.Fetch):
+        limit = limit_clause.args.get("count").this
+    else:  # dont have a limit or fetch clause
```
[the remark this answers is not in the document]
jczhong84: just not sure if there is anything else also treated as a limit clause, other than limit and fetch.

querybook/server/lib/query_analysis/transform.py
```diff
@@ -0,0 +1,169 @@
...
+    It returns a new statement with the limit applied and the original statement is not modified.
+    """
+    current_limit = get_select_statement_limit(statement_ast)
+    if current_limit is None or current_limit > 0:
+        return statement_ast
+
+    return statement_ast.limit(limit)
+
+
+def has_query_contains_unlimited_select(query: str, language: str) -> bool:
```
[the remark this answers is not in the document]
jczhong84: this is an equivalent function from the current js implementation. can revisit this if it's actually needed when we all switch to use this backend limit logic.

querybook/webapp/components/QueryRunButton/QueryRunButton.tsx
jczhong84: will address it in a separate PR to use backend limit logic.

querybook/webapp/components/QueryRunButton/QueryRunButton.tsx
jczhong84: will address it in a separate PR to use backend limit logic.

czgu (dismissed):

czgu (approved):

PR #1422, opened 2024-03-09
feat: editor awareness of default schemas

alexgann:
Video Demo/Explanation of Feature: https://youtu.be/w0z4pU_E0bM

### Issue
QueryBook QueryEditor is unaware of QueryEngine Connection URI's Default/Named database/schema

### Resolution (see video for walkthrough):
Add simple get_default_schema method to QueryEngine python model to parse any provided default schema on the connection URI
Make this defaultSchema known to the front-end for Linting, Analysis, and AutoCompletion
Retain support for use statements and explicit schemas when needed, and no-op to implicit "default" schema name when not provided

### DIFF
--- a/querybook/webapp/lib/sql-helper/sql-autocompleter.ts
+++ b/querybook/webapp/lib/sql-helper/sql-autocompleter.ts
@@ -240,9 +240,12 @@ export class SqlAutoCompleter {
             return [];
         }
 
+        let omit_schema = this.codeAnalysis.lineage.activeSchema || 'default';
+
         const { data: names } = await SearchTableResource.suggest(
             metastoreId,
-            prefix
+            prefix,
+            omit_schema
         );
         return names;
     }
--- a/querybook/webapp/lib/sql-helper/sql-lexer.ts
+++ b/querybook/webapp/lib/sql-helper/sql-lexer.ts
@@ -65,6 +65,7 @@ const contextSensitiveKeyWord = {
     table: 'table',
     update: 'table',
     join: 'table',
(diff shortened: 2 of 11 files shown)


alexgann:
@jczhong84 thanks again for taking a look!

querybook/webapp/lib/sql-helper/sql-lexer.ts
```diff
@@ -807,6 +813,7 @@ export function findTableReferenceAndAlias(statements: IToken[][]) {
     return {
         references,
         aliases,
+        activeSchema,
```
jczhong84, line 816: looks like `activeSchema` is just the same as `defaultSchema`, you dont need to define a new variable, but just return the defaultSchema.

querybook/webapp/lib/sql-helper/sql-autocompleter.ts
```diff
@@ -240,9 +240,12 @@ export class SqlAutoCompleter {
             return [];
         }
 
+        let omit_schema = this.codeAnalysis.lineage.activeSchema || 'default';
```
jczhong84, line 243: `const`.  also `|| default` is not necessarily needed, as the API has the default value `default`  for the 3rd parameter.  could be just below
```
suggest(
            metastoreId,
            prefix,
            this.codeAnalysis.lineage.activeSchema)
```

jczhong84:
@alexgann sorry for not looking it sooner. thanks for the demo and PR! this will be quite useful. left some comments

PR #1423, opened 2024-03-11
fix: Metastore ACL support for wildcard prefix match

baumandm:
Adds support for wildcard prefixes on table names in the ACL rules.  E.g. `querybook2.query_*` matches any tables in the `querybook2` schema starting with `query_`.

Also added a helper tooltip to the admin UI to document the supported formats, since I don't think there's any other documentation on how to use this feature.

Screenshot 2024-03-11 at 3 14 50 PM


czgu (approved):
thanks!

PR #1424, opened 2024-03-11
fix: added sort key when more data tables are loaded in a search, causing results to be sorted correctly

jij1949:
This PR fixes an issue where the table results in a search would not sort correctly after the first 100 results were loaded.


jczhong84 (approved):

jczhong84:
@jij1949 thanks for the fix!

PR #1430, opened 2024-04-03
fix: add bom to downloaded csv for excel

rchandnaeg:
If you download the result of a query as a CSV and then open the result in Excel, there are encoding issues for unicode/non-latin characters. This PR makes the process of opening a downloaded CSV in Excel a little smoother by adding the Byte Order Mark to the downloaded CSV. There is not a pressing issue with Querybook right now, Excel is just very particular and this makes the downloaded CSVs play nice with Excel's strange encoding import rules.

Example of Issue: Upload the following table and run `SELECT *`, then download the results of the query execution 

[non-latin-test.csv](https://github.com/pinterest/querybook/files/14858655/non-latin-test.csv)

Opening the result CSV in Excel used to give the following:
image

This PR adds the BOM so that Excel is happy and gets the right encoding:
image


jczhong84:
@rchandnaeg thanks for the PR.

I'm not sure if it is the best way to fix this issue, as some other applications may have issue when the file has BOM. I think a better solution might be to provide a separate option with the BOM, or when user downloads, they have an option to add the bom for ms excel?

baumandm:
@jczhong84 what do you think about an XLXS export option to avoid this (and other) CSV import issues into Excel?

jczhong84:
xlxs export option sounds good to me

PatersonKopp:
> xlxs export option sounds good to me

Please check out https://github.com/pinterest/querybook/pull/1557, we also needed it

PR #1432, opened 2024-04-06
feat: add @mention to select a table for the command input

jczhong84:
- use "@" to mention what table to use
- if it's a new cell, it will populate a simple select * query for the first mentioned table automatically
- use cmd+/ to reset and open the commands

Also updated the table suggest API to cover non prefix suggestion.

package used: https://github.com/signavio/react-mentions

text2sql-table

### DIFF
--- a/querybook/server/lib/elasticsearch/suggest_table.py
+++ b/querybook/server/lib/elasticsearch/suggest_table.py
@@ -1,17 +1,18 @@
 def construct_suggest_table_query(
-    prefix: str,
+    keyword: str,
     limit: int,
     metastore_id: int,
 ):
     return {
-        "suggest": {
-            "suggest": {
-                "text": prefix,
-                "completion": {
-                    "field": "completion_name",
-                    "size": limit,
-                    "contexts": {"metastore_id": metastore_id},
+        "from": 0,
+        "size": limit,
+        "query": {
+            "bool": {
+                "must": {
+                    "match": {"full_name_ngram": {"query": keyword, "operator": "and"}}
                 },
+                "filter": {"match": {"metastore_id": metastore_id}},
             }
         },
+        "_source": ["id", "full_name"],
     }
--- a/querybook/webapp/components/AIAssistant/AICommandBar.tsx
+++ b/querybook/webapp/components/AIAssistant/AICommandBar.tsx
@@ -100,9 +100,10 @@ export const AICommandBar: React.FC<IQueryCellCommandBarProps> = forwardRef(
         useEffect(() => {
             if (!query && mentionedTables.length === 1) {
                 onUpdateQuery(
-                    `SELECT\n  *\nFROM\n  ${mentionedTables[0]}\nLIMIT\n  10`,
+                    `SELECT * FROM ${mentionedTables[0]} LIMIT 10`,
                     false
                 );
+                onFormatQuery();
             }
         }, [mentionedTables]);
 
--- a/querybook/webapp/components/AIAssistant/AICommandBar.tsx
+++ b/querybook/webapp/components/AIAssistant/AICommandBar.tsx
@@ -1,5 +1,11 @@
 import { uniq } from 'lodash';
-import React, { forwardRef, useCallback, useEffect, useState } from 'react';
+import React, {
+    forwardRef,
+    useCallback,
+    useEffect,
+    useMemo,
+    useState,
+} from 'react';
 
 import { AICommandInput } from 'components/AIAssistant/AICommandInput';
 import { AICommandResultView } from 'components/AIAssistant/AICommandResultView';
@@ -62,8 +68,8 @@ export const AICommandBar: React.FC<IQueryCellCommandBarProps> = forwardRef(
             (cmd) => cmd.name === (query ? 'edit' : 'generate')
         );
         const tablesInQuery = useTablesInQuery(query, queryEngine.language);
-        const [tables, setTables] = useState(tablesInQuery);
-        const [mentionedTables, setMentionedTables] = useState([]);
+        const [tables, setTables] = useState<string[]>([]);
+        const [mentionedTables, setMentionedTables] = useState<string[]>([]);
         const commandInputRef = useForwardedRef<IResizableTextareaHandles>(ref);
         const [showPopupView, setShowPopupView] = useState(false);
         const [command, setCommand] =
@@ -75,6 +81,14 @@ export const AICommandBar: React.FC<IQueryCellCommandBarProps> = forwardRef(
         );
         const [showConfirm, setShowConfirm] = useState(false);
 
+        const finalTablesToUse = useMemo(() => {
+            if (mentionedTables.length > 0) {
+                return mentionedTables;
+            } else {
+                return uniq([...tablesInQuery, ...tables]);
+            }
+        }, [tablesInQuery, mentionedTables, tables]);
+
         const {
             runCommand,
             isRunning,
@@ -82,16 +96,6 @@ export const AICommandBar: React.FC<IQueryCellCommandBarProps> = forwardRef(
             commandResult,
             resetCommandResult,
         } = useCommand(command, commandRunner);
-
-        useEffect(() => {
-            setTables((tables) => {
-                if (mentionedTables.length > 0) {
-                    return mentionedTables;
-                } else {
-                    return uniq([...tablesInQuery, ...tables]);
-                }
-            });
-        }, [tablesInQuery, mentionedTables]);
 
         useEffect(() => {
             if (!query && mentionedTables.length === 1) {
@@ -109,7 +113,7 @@ export const AICommandBar: React.FC<IQueryCellCommandBarProps> = forwardRef(
             } else if (command.name === 'generate' || command.name === 'edit') {
                 setCommandKwargs({
                     query_engine_id: queryEngine.id,
-                    tables: tables,
+                    tables: finalTablesToUse,
                     question: commandInputValue,
                     original_query: command.name === 'generate' ? '' : query,
                 });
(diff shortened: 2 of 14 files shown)


querybook/webapp/components/AIAssistant/AICommandBar.tsx
```diff
@@ -86,8 +84,23 @@ export const AICommandBar: React.FC<IQueryCellCommandBarProps> = forwardRef(
...
+                } else {
+                    return uniq([...tablesInQuery, ...tables]);
+                }
+            });
+        }, [tablesInQuery, mentionedTables]);
+
+        useEffect(() => {
+            if (!query && mentionedTables.length === 1) {
+                onUpdateQuery(
+                    `SELECT\n  *\nFROM\n  ${mentionedTables[0]}\nLIMIT\n  10`,
```
czgu, line 99: its easier to do `formatQuery("SELECT * FROM table LIMIT 10")`

querybook/webapp/components/AIAssistant/AICommandBar.tsx
```diff
@@ -66,6 +63,7 @@ export const AICommandBar: React.FC<IQueryCellCommandBarProps> = forwardRef(
         );
         const tablesInQuery = useTablesInQuery(query, queryEngine.language);
         const [tables, setTables] = useState(tablesInQuery);
+        const [mentionedTables, setMentionedTables] = useState([]);
```
czgu, line 66: typing?

querybook/webapp/components/AIAssistant/AICommandBar.tsx
```diff
@@ -86,8 +84,23 @@ export const AICommandBar: React.FC<IQueryCellCommandBarProps> = forwardRef(
...
-            setTables((tables) => uniq([...tablesInQuery, ...tables]));
-        }, [tablesInQuery]);
+            setTables((tables) => {
+                if (mentionedTables.length > 0) {
+                    return mentionedTables;
+                } else {
+                    return uniq([...tablesInQuery, ...tables]);
+                }
+            });
+        }, [tablesInQuery, mentionedTables]);
```
czgu, line 94: too many mutation dependencies, how about we start from a fresh slate with a useMemo?

querybook/webapp/components/AIAssistant/AICommandInput.tsx
```diff
@@ -139,6 +196,33 @@ export const AICommandInput: React.FC<AICommandInputProps> = forwardRef(
             </div>
         );
 
+        const loadTables = useCallback(
+            (keywords: string, callback) => {
```
czgu, line 200: typing for callback?

querybook/webapp/components/AIAssistant/AICommandInput.tsx
```diff
@@ -139,6 +196,33 @@ export const AICommandInput: React.FC<AICommandInputProps> = forwardRef(
...
+                    return;
+                }
+
+                SearchTableResource.searchConcise({
+                    metastore_id: metastoreId,
+                    keywords,
+                }).then(({ data }) => {
+                    const filteredTableNames = data.results.filter(
+                        (result) =>
+                            tables.indexOf(
```
czgu, line 211: why not use .includes?

querybook/webapp/const/keyMap.ts
```diff
@@ -85,6 +85,12 @@ const DEFAULT_KEY_MAP = {
             name: 'Delete current cell',
         },
     },
+    commandBar: {
```
czgu, line 88: AICommandBar

querybook/webapp/components/AIAssistant/AICommandBar.tsx
```diff
@@ -226,6 +239,9 @@ export const AICommandBar: React.FC<IQueryCellCommandBarProps> = forwardRef(
                             );
                             setCommandInputValue(inputValue);
                         }}
+                        tables={mentionedTables}
+                        onTablesChange={setMentionedTables}
```
czgu, line 243: rename prop to mentionedTables?

querybook/webapp/components/AIAssistant/AICommandInput.tsx
```diff
@@ -80,46 +131,52 @@ export const AICommandInput: React.FC<AICommandInputProps> = forwardRef(
...
         );
 
         const onKeyDown = useCallback(
             (event: React.KeyboardEvent) => {
-                if (filteredCommands.length > 0) {
+                let handled = true;
+                // Cmd + / will reset the command and value and open the command options
+                if (matchKeyMap(event, KeyMap.commandBar.openCommands)) {
+                    setNewCommand(undefined);
+                    setCommandValue('/');
```
czgu, line 146: seems like a pretty dangerous shortcut that would delete the existing text

querybook/webapp/components/AIAssistant/AICommandBar.tsx
jczhong84: formatQuery doesn't return anything, but will change the editor content in-place

querybook/webapp/components/AIAssistant/AICommandInput.tsx
jczhong84: the reason I added this is when to switch commands, it's quite cumbersome to delete all the text and then backspace the command. I guess it's not that easy for user to press this shortcut? or we can think of adding undo in the future.

querybook/webapp/components/AIAssistant/AICommandBar.tsx
czgu: use `import { format } from 'lib/sql-helper/sql-formatter';`

querybook/webapp/components/AIAssistant/AICommandInput.tsx
```diff
@@ -139,6 +196,38 @@ export const AICommandInput: React.FC<AICommandInputProps> = forwardRef(
...
+                keywords: string,
+                callback: (
+                    options: Array<{ id: string; display: string }>
+                ) => void
+            ) => {
+                if (!keywords) {
+                    return;
+                }
+
+                SearchTableResource.searchConcise({
```
czgu, line 210: I tried this out and noticed the first 2 letters doesn't return any responses, it would make this feature hard to discover for our users

querybook/webapp/components/AIAssistant/AICommandInput.tsx
```diff
@@ -153,21 +242,35 @@ export const AICommandInput: React.FC<AICommandInputProps> = forwardRef(
...
                     value={commandValue}
                     onChange={handleChange}
-                    className="question-text-area"
+                    onKeyDown={onKeyDown}
+                    style={mentionInputStyle}
                     placeholder={
                         command
                             ? command.hint
-                            : 'Ask AI to generate/edit the query, or type / to see more options'
+                            : 'Ask AI to generate/edit the query. Type @ to select a table. Type / to see more commands. Type Cmd+/ to reset the command.'
```
czgu, line 253: instead of hard coding Cmd+/, can you use KeyMap.aiCommandBar.openCommands instead

querybook/webapp/components/AIAssistant/AICommandInput.tsx
jczhong84: this should be some issue of my devapp index, it is supposed to start searching when people type at least one char.

querybook/webapp/components/AIAssistant/AICommandBar.tsx
jczhong84: calling onFormatQuery after updating query instead

querybook/webapp/components/AIAssistant/AICommandInput.tsx
czgu: wait I also noticed we should be using SearchTableResource.suggest instead of search

querybook/webapp/components/AIAssistant/AICommandInput.tsx
```diff
@@ -139,6 +196,38 @@ export const AICommandInput: React.FC<AICommandInputProps> = forwardRef(
...
+                    keywords,
+                }).then(({ data }) => {
+                    const filteredTableNames = data.results.filter(
+                        (result) =>
+                            !mentionedTables.includes(
+                                `${result.schema}.${result.name}`
+                            )
+                    );
+                    const tableNameOptions = filteredTableNames.map(
+                        ({ id, schema, name }) => ({
```
czgu, line 221: id is not used here?

czgu: i originally thought it would be used in line 222, but it would have been `[id]: ...`

querybook/webapp/components/AIAssistant/AICommandInput.tsx
jczhong84: this piece of code was copied over from somewhere, didn't check them :( . let me revise them

jczhong84: just realized that `.suggest` only works for prefix, I think `.searchConcise` might be more useful here, which can cover any word inside the table name.

querybook/webapp/components/AIAssistant/AICommandInput.tsx
jczhong84: was meant to use it, but found not necessary, forgot to remove it.

querybook/server/lib/elasticsearch/suggest_table.py
```diff
@@ -1,17 +1,18 @@
...
+    keyword: str,
     limit: int,
     metastore_id: int,
 ):
     return {
-        "suggest": {
-            "suggest": {
-                "text": prefix,
-                "completion": {
-                    "field": "completion_name",
```
czgu, line 11: since completion name is getting removed, can you add some comment in the elasticsearch definition on that field so we know to delete it for future? Also I guess we can stop indexing any data to that field

czgu (approved):

PR #1436, opened 2024-04-11
feat: Add current_user variable and slugify filter to Jinja

baumandm:
This PR adds two new Jinja predefined variables, as well as a custom `slugify` filter using the [`python-slugify`](https://pypi.org/project/python-slugify/) package.

- `{{ current_user }}`: Renders the current user's username
- `{{ current_user_email }}`: Renders the current user's email address
- `{{ ... | slugify }}`: Transforms the input into a slugified version with underscores, stripping spaces, Unicode, and special characters.  This is suitable for generating table / column names dynamically.

Screenshot 2024-04-11 at 09-13-58 Additional Jinja Variables _ Filters - Querybook
Screenshot 2024-04-11 at 9 11 57 AM
Screenshot 2024-04-11 at 9 12 04 AM


querybook/server/lib/query_analysis/templating.py
```diff
@@ -315,7 +326,7 @@ def get_templated_query_variables(variables_provided, jinja_env):
 
 
 def render_templated_query(
-    query: str, variables: Dict[str, str], engine_id: int, session=None
+    query: str, variables: Dict[str, str], engine_id: int, uid: int, session=None
```
czgu, line 329:
```suggestion
    query: str, variables: Dict[str, str], engine_id: int, uid: Optional[int] = None, session=None
```

querybook/server/lib/query_analysis/templating.py
```diff
@@ -164,12 +168,19 @@ def get_latest_partition(
     return get_latest_partition
 
 
-def get_templated_query_env(engine_id: int, session=None):
+def get_templated_query_env(engine_id: int, user: User, session=None):
```
czgu, line 171:
```suggestion
def get_templated_query_env(engine_id: int, user: Optional[User], session=None):
```

czgu (commented):
looks great

PR #1437, opened 2024-04-11
fix: Fix email notifications to multiple email addresses

baumandm:
This fixes an issue with sending notifications to multiple email addresses (not users) at once.  The first email address would get the email, and the email would look like it was sent to all recipients, but it would not actually be delivered to the other email addresses.  `smtp.sendmail()` needs a list of addresses rather than the joined string in the message header.

This doesn't apply to any Querybook users who are added to the notification, because they are notified individually.

This StackOverflow helps explain the issue and the fix: https://stackoverflow.com/q/8856117.

Screenshot 2024-04-11 at 2 48 08 PM


jczhong84 (approved):
thanks for the fix! @baumandm

PR #1438, opened 2024-04-12
fix: revert to use searchConcise for mentioning a table

jczhong84:
in the PR https://github.com/pinterest/querybook/pull/1432/, we updated the table suggest api, which makes the performance really bad as it has no weight. this PR is to revert it and use `searchConcise` for the table mention instead.

Also fix and refactor a bit for the tables management in the AICommandBar component.

### DIFF
--- a/querybook/webapp/components/AIAssistant/AICommandBar.tsx
+++ b/querybook/webapp/components/AIAssistant/AICommandBar.tsx
@@ -19,8 +19,6 @@ import { IQueryEngine } from 'const/queryEngine';
 import { CommandRunner, useCommand } from 'hooks/useCommand';
 import { useForwardedRef } from 'hooks/useForwardedRef';
 import { trackClick } from 'lib/analytics';
-import { TableToken } from 'lib/sql-helper/sql-lexer';
-import { analyzeCode } from 'lib/web-worker';
 import { Button } from 'ui/Button/Button';
 import { Message } from 'ui/Message/Message';
 import { IResizableTextareaHandles } from 'ui/ResizableTextArea/ResizableTextArea';
--- a/querybook/webapp/components/AIAssistant/AICommandBar.tsx
+++ b/querybook/webapp/components/AIAssistant/AICommandBar.tsx
@@ -38,28 +38,6 @@ interface IQueryCellCommandBarProps {
     ref: React.Ref<IResizableTextareaHandles>;
 }
 
-// const useTablesInQuery = (query: string, language: string) => {
-//     const [tables, setTables] = useState<string[]>([]);
-
-//     useEffect(() => {
-//         if (!query) {
-//             return;
-//         }
-
-//         analyzeCode(query, 'autocomplete', language).then((codeAnalysis) => {
-//             const tableReferences: TableToken[] = [].concat.apply(
-//                 [],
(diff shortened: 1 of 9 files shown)


querybook/webapp/components/AIAssistant/AICommandBar.tsx
```diff
@@ -81,13 +63,25 @@ export const AICommandBar: React.FC<IQueryCellCommandBarProps> = forwardRef(
...
+        /**
+         * There are three types of tables:
+         * 1. Tables used in the query.
+         * 2. Tables mentioned in the command.
+         * 3. Tables selected in the table selector.
+         *
+         * Regarding which tables to use:
+         * - If there are mentioned tables in the command, use those.
+         * - If there are no mentioned tables, use the tables in the query (type 1) merged with the selected tables (type 3).
+         * - However, selected tables will override the tables in the query. If a table is deselected, it will be removed from the list, including the tables in the query.
```
czgu, line 75: even if it is deselected from tables, isn't it still in tablesInQuery

querybook/webapp/components/DataDocQueryCell/DataDocQueryCell.tsx
```diff
@@ -124,6 +124,7 @@ interface IState {
     showRenderedTemplateModal: boolean;
     showUDFModal: boolean;
     hasLintError: boolean;
+    tables: Array<string>;
```
czgu, line 127: nit: string[]

czgu: nit: tableNamesInQuery: string[]

querybook/webapp/components/DataDocQueryCell/DataDocQueryCell.tsx
```diff
@@ -702,7 +704,10 @@ class DataDocQueryCellComponent extends React.PureComponent<IProps, IState> {
                 };
             }
         });
-        this.setState({ samplingTables });
+        this.setState({
```
czgu, line 707: do you recall why this function is async but there is no await?

querybook/webapp/components/AIAssistant/AICommandBar.tsx
jczhong84: yes, it is still in `tablesInQuery`, but will not be in the `tables` state, which is connected with the multi table selector. so the table selector can add/remove any table, including `tablesInQuery`. 

Table in the query will only be merged into `tables` when `tablesInQuery` changes.

querybook/webapp/components/DataDocQueryCell/DataDocQueryCell.tsx
jczhong84: probably it was implemented by getting the tables info in this function and somehow I changed it and forgot to remove the async :)

querybook/webapp/components/AIAssistant/AICommandBar.tsx
czgu: oh I got it, this is to reset tables with tablesInQuery whenever tablesInQueryChanges. 👍

czgu (approved):

PR #1450, opened 2024-05-13
feat: Add any links associated with a table to querybook descriptions

zhangvi7:
### Description
- Tables from metatore maybe have links associated to table descriptions
- Exposing any links that are attached to tables in Querybook
- Adding `table_links` column to the data_table_information table and load it from metastore.
### Test Plan
image


querybook/migrations/versions/767cf0f573bd_add_table_links_to_datatable.py
```diff
@@ -0,0 +1,84 @@
+"""add table links to datatable
```
kgopal492, line 1: You should probably edit this file to only include the changes made in this PR (aka add column table_links for upgrade and drop_column table_links for downgrade)

querybook/webapp/const/metastore.ts
```diff
@@ -74,7 +74,7 @@ export interface IDataTable {
         partition_keys?: string[];
     };
     custom_properties?: Record<string, string | number>;
-
+    table_links?: Record<string, string>[];
```
jczhong84, line 77: it looks like a link can only have `label` and `url`, can we define a new type for it with `label` and `url`, while `label` as optional?

querybook/webapp/components/DataTableViewOverview/DataTableViewOverview.tsx
```diff
@@ -130,6 +130,16 @@ export const DataTableViewOverview: React.FC<
         />
     ) : null;
 
+    const tableLinks = table.table_links ?? [];
+    const tableLinksDOM = tableLinks.map((link, index) => (
```
jczhong84, line 134: nit: seems unnecessary to introduce a new variable, just `(table.table_links ?? []).map`

querybook/server/const/metastore.py
```diff
@@ -75,6 +75,9 @@ class DataTable(NamedTuple):
     # Custom properties
     custom_properties: dict[str, str] = None
 
+    # Links associated to table
+    table_links: list[dict[str, str]] = None
```
czgu, line 79: could we use typed dict here?

querybook/webapp/const/metastore.ts
czgu: +1, using Record is not clear

zhangvi7:
- fixed lint
- addressed comments

czgu (approved):

PR #1453, opened 2024-05-17
ui: Add icon to table links

zhangvi7:
- Add link icon to table links to make link attachments more obvious
Screenshot 2024-05-17 at 5 11 52 PM

### DIFF
--- a/querybook/webapp/components/DataTableViewOverview/DataTableViewOverview.tsx
+++ b/querybook/webapp/components/DataTableViewOverview/DataTableViewOverview.tsx
@@ -41,6 +41,7 @@ import { Message } from 'ui/Message/Message';
 import { ShowMoreText } from 'ui/ShowMoreText/ShowMoreText';
 
 import { DataTableViewOverviewSection } from './DataTableViewOverviewSection';
+import { Link2 } from 'lucide-react';
 
 import './DataTableViewOverview.scss';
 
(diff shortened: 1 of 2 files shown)


querybook/webapp/components/DataTableViewOverview/DataTableViewOverview.tsx
```diff
@@ -133,6 +134,14 @@ export const DataTableViewOverview: React.FC<
     const tableLinksDOM = (table.table_links ?? []).map((link, index) => (
         <div key={index}>
             <Link to={link.url} newTab>
+                <Link2
+                    size={20}
+                    style={{
+                        position: 'relative',
+                        top: '5px',
+                        paddingRight: '5px',
+                    }}
```
jczhong84, line 143: can you create a css class for it?

also, instead of using relative position, would it better to use a inline flexbox on the parent Link? like

{ display: 'inline-flex', alignItems: 'center', gap: '5px' }.

btw, it might be better to use times of 4px for all spaces

czgu: +1, we are enforcing padding in multiples of 4px

querybook/webapp/components/DataTableViewOverview/DataTableViewOverview.tsx
```diff
@@ -41,6 +41,7 @@ import { Message } from 'ui/Message/Message';
 import { ShowMoreText } from 'ui/ShowMoreText/ShowMoreText';
 
 import { DataTableViewOverviewSection } from './DataTableViewOverviewSection';
+import { Link2 } from 'lucide-react';
```
czgu, line 44: use Icon component for icons

jczhong84 (approved):

PR #1459, opened 2024-06-12
chore: log sample rate in query run context logs

jczhong84:


querybook/webapp/components/DataDocQueryCell/DataDocQueryCell.tsx
```diff
@@ -467,6 +467,7 @@ class DataDocQueryCellComponent extends React.PureComponent<IProps, IState> {
             element: ElementType.RUN_QUERY_BUTTON,
             aux: {
                 lintError: this.state.hasLintError,
+                sampleRate: this.sampleRate,
```
czgu, line 470: could we only log sampleRate when sample rate is actually used? there could be a case where user selects sampling table and a rate, and then switch the table with no sample right?

querybook/webapp/components/DataDocQueryCell/DataDocQueryCell.tsx
```diff
@@ -232,6 +228,11 @@ class DataDocQueryCellComponent extends React.PureComponent<IProps, IState> {
         return Object.keys(this.samplingTables).length > 0;
     }
 
+    public get sampleRate() {
+        // -1 or tables don't support sampling, 0 for sample rate is none
```
czgu, line 232: typo `-1 for`

czgu: // -1 for tables don't support sampling, 0 for default sample rate (which means disable sampling)

czgu (approved):

querybook/webapp/components/DataDocQueryCell/DataDocQueryCell.tsx
```diff
@@ -232,6 +228,11 @@ class DataDocQueryCellComponent extends React.PureComponent<IProps, IState> {
         return Object.keys(this.samplingTables).length > 0;
     }
 
+    public get sampleRate() {
+        // -1 for tables don't support sampling, 0 for default sample rate (which means disable sampling)
+        return this.hasSamplingTables ? this.state.meta.sample_rate ?? 0 : -1;
```
zhangvi7, line 233: @jczhong84  this cause max call stack size exceed since sampleRate() calls sampleRate() which calls this.sampleRate() again, I can revert it in my sampling PR if that sounds good to you

Issue #1463, opened 2024-06-18
Querybook fails to start up - Numpy update


kousiksundara:
Numpy released a new version a couple days ago: https://github.com/numpy/numpy/releases/tag/v2.0.0
As part of it, it has made breaking changes `Alias np.float_ has been removed. Use np.float64 instead.`

Building the dockerfile and running causes the following error:
`INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
ERROR [alembic.util.messaging] Can't locate revision identified by '8996bf11227d'
FAILED: Can't locate revision identified by '8996bf11227d'
Traceback (most recent call last):
  File "/app/querybook/server/scripts/init_es.py", line 1, in <module>
    from logic.elasticsearch import create_indices_if_not_exist
  File "/app/querybook/server/logic/elasticsearch.py", line 13, in <module>
    from lib.elasticsearch.search_utils import (
  File "/app/querybook/server/lib/elasticsearch/search_utils.py", line 3, in <module>
    from elasticsearch import Elasticsearch, RequestsHttpConnection
  File "/app/.local/lib/python3.9/site-packages/elasticsearch/__init__.py", line 36, in <module>
    from .client import Elasticsearch
  File "/app/.local/lib/python3.9/site-packages/elasticsearch/client/__init__.py", line 23, in <module>
    from ..transport import Transport, TransportError
  File "/app/.local/lib/python3.9/site-packages/elasticsearch/transport.py", line 31, in <module>
    from .serializer import DEFAULT_SERIALIZERS, Deserializer, JSONSerializer
  File "/app/.local/lib/python3.9/site-packages/elasticsearch/serializer.py", line 50, in <module>
    np.float_,
  File "/app/.local/lib/python3.9/site-packages/numpy/__init__.py", line 397, in __getattr__
    raise AttributeError(
AttributeError: `np.float_` was removed in the NumPy 2.0 release. Use `np.float64` instead.
/app
Running: python3 querybook/server/runweb.py --debug --webpack 10001
Traceback (most recent call last):
  File "/app/querybook/server/runweb.py", line 22, in <module>
    from app.server import flask_app
  File "/app/querybook/server/app/server.py", line 14, in <module>
    import datasources
  File "/app/querybook/server/datasources/__init__.py", line 1, in <module>
    from . import admin
  File "/app/querybook/server/datasources/admin.py", line 13, in <module>
[...]

PR #1466, opened 2024-06-24
fix: Pin NumPy version to <2.0.0 to avoid breaking changes

Ari1009:
This change pins the NumPy version to <2.0.0 to avoid breaking changes caused by the removal of the `np.float_` alias in NumPy 2.0.0. This ensures compatibility with packages that still rely on the deprecated alias.
Fixes- #1463


czgu (approved):
Thank you!

PR #1472, opened 2024-06-30
docs: documentation typo https

Ari1009:
Visit http://localhost:10001, Small fix correct URL is http://localhost:10001, not https://localhost:10001.


Ari1009:
@czgu can you take a look, thanks!!

czgu (approved):
thanks for the update!

PR #1473, opened 2024-07-03
feat: pass api_access_token to executor client

zhangvi7:
If we use `api-access-token` in request header, then pass it down to the executor client

Then in Presto client, we can use this flag to determine whether to continue using ldap auth or pass outh jwt instead.

### DIFF
--- a/querybook/server/datasources/query_execution.py
+++ b/querybook/server/datasources/query_execution.py
@@ -87,9 +87,10 @@ def create_query_execution(
 
         try:
             run_query_task.apply_async(
-                args=[
-                    query_execution.id,
-                ]
+                kwargs={
+                    "query_execution_id": query_execution.id,
+                    "api_access_token": current_user.api_access_token,
+                },
             )
             query_execution_dict = query_execution.to_dict()
 
--- a/querybook/server/lib/query_executor/base_executor.py
+++ b/querybook/server/lib/query_executor/base_executor.py
@@ -502,10 +502,12 @@ class QueryExecutorBaseClass(metaclass=ABCMeta):
         statement_ranges,
         client_setting,
         execution_type,
+        api_access_token=False,
     ):
         self._query = query
         self._query_execution_id = query_execution_id
         self._execution_type = execution_type
+        self._api_access_token = api_access_token
 
         if self.SINGLE_QUERY_QUERY_ENGINE():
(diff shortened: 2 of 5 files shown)


zhangvi7:
- Use execution metadata instead to hold access token 
- Manually tested with api access token
image

querybook/server/datasources/query_execution.py
```diff
@@ -72,10 +72,14 @@ def create_query_execution(
...
-                query_execution.id, metadata, session=session
-            )
+        api_access_token = (
+            True if request.headers.get("api-access-token", None) else False
+        )
+        metadata = metadata or {}
+        metadata["api_access_token"] = api_access_token
+        logic.create_query_execution_metadata(
+            query_execution.id, metadata, session=session
+        )
```
czgu, line 82:
```
metadata = {}
used_api_token = request.headers.get("api-access-token", None) is not None
if used_api_token:
    metadata["used_api_token"] = used_api_token

if metadata:
   logic.create(...)
```

querybook/server/lib/query_executor/base_executor.py
```diff
@@ -528,6 +528,16 @@ def __init__(
         self._client = None
         self._cursor = None
 
+        with DBSession() as session:
```
czgu, line 531: dbsession is not needed if you are only calling once

querybook/server/lib/query_executor/base_executor.py
```diff
@@ -528,6 +528,16 @@ def __init__(
...
+            query_execution_metadata = (
```
czgu, line 532: this might be None? (if metadata wasn't created)

querybook/server/lib/query_executor/base_executor.py
```diff
@@ -528,6 +528,16 @@ def __init__(
...
+                qe_logic.get_query_execution_metadata_by_execution_id(
+                    self._query_execution_id, session=session
+                ).execution_metadata
+            )
+            self._api_access_token = query_execution_metadata.get(
```
czgu, line 537: rename to used_api_token or some other boolean name

api_access_token was used as a str to represent the actual token, let's not reuse the same name for a boolean flag?

czgu (approved):

PR #1482, opened 2024-08-02
fix: Metastore sync tables with a single transaction

baumandm:
We were facing some database perf issues, so I looked into it and the SQL query with the most waits was `INSERT INTO 'data_table_warnings'`:

Screenshot 2024-07-12 at 10 37 24 AM

We recently updated our metastore loader code to automatically add warnings to certain tables during the sync.  That explains why we're seeing it now, but it didn't seem right so I looked into it.

While most of the metastore sync executes as a single transaction per table, I realized data table warnings are being committed immediately.  I enabled database logging in SQLAlchemy and ran through the metastore sync, which helped identify a few other areas where separate transactions were being committed.

This PR fixes every case of a separate transaction in the `base_metastore_loader`'s `_create_table_table` function that I could detect.

1. Where necessary, I added a `commit` arg to functions that were missing it (and passing `commit=False` during the sync)
2. Functions that have a `commit` arg should always pass `commit=False` to all sub-query functions, since they (optionally) commit at the end of the function
 
With these changes, each table creates a single transaction that is committed (or rolled back) at the completion of the `_create_table_table()` function.

We deployed this change to production late on July 30th, and the load dropped significantly; all the `row_lock_waits` are gone:

Screenshot 2024-08-02 at 10 33 16 AM

Moreover, the metastore loader task runs 100% faster now:

Screenshot 2024-08-02 at 5 19 24 PM

Note: to enable database logging, add `echo=True` to the `create_engine()` call in `db.py`.  This let me know exactly when transactions were being committed.

### DIFF
--- a/querybook/server/lib/metastore/base_metastore_loader.py
+++ b/querybook/server/lib/metastore/base_metastore_loader.py
@@ -400,7 +400,10 @@ class BaseMetastoreLoader(metaclass=ABCMeta):
                 )
 
             delete_column_not_in_metastore(
-                table_id, set(map(lambda c: c.name, columns)), session=session
+                table_id,
+                set(map(lambda c: c.name, columns)),
+                commit=False,
+                session=session,
             )
 
             for column in columns:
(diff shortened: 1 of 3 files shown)


czgu (approved):
Nice work!

PR #1489, opened 2024-08-23
fix: str_to_bool and sql-limiter

jczhong84:
- str_to_bool: remove the bool input type
 - sql-limiter: `parsedStatement` could be undefined

### DIFF
--- a/querybook/server/lib/utils/utils.py
+++ b/querybook/server/lib/utils/utils.py
@@ -162,5 +162,5 @@ def get_default_args(func):
     }
 
 
-def str_to_bool(value: Union[str, bool]):
-    return value.lower() in (True, "yes", "true", "t", "1")
+def str_to_bool(value: str):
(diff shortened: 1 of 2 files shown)


querybook/server/lib/utils/utils.py
```diff
@@ -162,5 +162,5 @@ def get_default_args(func):
     }
 
 
-def str_to_bool(value: Union[str, bool]):
-    return value.lower() in (True, "yes", "true", "t", "1")
+def str_to_bool(value: str):
+    return False if value is None else value.lower() in ("yes", "true", "t", "1")
```
czgu, line 166:
```suggestion
def str_to_bool(value: Optional[str]):
    if value is None:
         return False
    return  value.lower() in ("yes", "true", "t", "1")
```

czgu (approved):

PR #1490, opened 2024-08-29
chore: add click event logs for export buttons

jczhong84:
add click event logs for export buttons and insert snippets


zhangvi7 (approved):
lgtm

PR #1519, opened 2024-11-19
Multiple Editor Style Changes

czgu:
- Updated monokai style so the default text is white instead of orange
- Added per language keyword config
- Altered indent rule for something simpler: it just indent based on the previous line indent level.


querybook/webapp/components/QueryEditor/QueryEditor.tsx
```diff
@@ -295,6 +295,7 @@ export const QueryEditor: React.FC<
             useHoverTooltipExtension({
                 codeAnalysisRef,
                 metastoreId: 1,
+                language,
```
jczhong84, line 298: can you help to update the hardcoded metastoreId above as well? thanks!

jczhong84 (approved):

PR #1521, opened 2024-11-21
feat: add table sample rate to user settings

jczhong84:
Add the user settings for the default table sample rate, so that when a new query cell is added, the prefered sample rate will be selected by default if the table supports sampling

Also adding some util functions for the public configs

sampling_settings

### DIFF
--- a/querybook/webapp/components/QueryRunButton/QueryRunButton.tsx
+++ b/querybook/webapp/components/QueryRunButton/QueryRunButton.tsx
@@ -303,18 +303,11 @@ const TableSamplingSelector: React.FC<{
     );
 
     React.useEffect(() => {
-        if (
-            sampleRate === undefined ||
-            !sampleRateOptions.some((option) => option.value === sampleRate)
-        ) {
+        // If it is a new cell without the sample rate selected, use the default sample rate from user settings
+        if (sampleRate === undefined) {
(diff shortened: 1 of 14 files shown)


querybook/webapp/components/QueryRunButton/QueryRunButton.tsx
```diff
@@ -287,29 +291,30 @@ const QueryLimitSelector: React.FC<{
...
+    );
+
     React.useEffect(() => {
-        if (!sampleRateOptions.some((option) => option.value === sampleRate)) {
-            setSampleRate(DEFAULT_SAMPLE_RATE);
+        if (
+            sampleRate === undefined ||
+            !sampleRateOptions.some((option) => option.value === sampleRate)
+        ) {
+            setSampleRate(parseFloat(userDefaultTableSampleRate));
```
czgu, line 310: do we need to watch out if userDefaultTableSampleRate is not in sampleRateOptions which could cause loop?

jczhong84: in theory, the sample rate options of the user settings(including the user selected one) or here are both from the public config file, what you mentioned can not happen. But I can add a check to select the first one if it happens.

czgu (approved):

PR #1526, opened 2024-12-06
feat: add column hover tooltip and column value auto completion

jczhong84:
* If the column stats has distinct values, it will show the auto completion for the values in the where clause
Snip20241205_21

* when hover over the column names, it will show a tooltip, like the table tooltip
image


querybook/webapp/lib/sql-helper/sql-parser.ts
```diff
@@ -0,0 +1,329 @@
...
+                (stat) => stat.key === 'distinct_values'
+            );
+            if (distinctValuesStat?.value instanceof Array) {
+                return distinctValuesStat?.value;
+            }
+
+            return [];
+        }
+
+        return [];
```
czgu, line 151:
```suggestion
        if (columns.length !== 1) return [];
         
       const colStats = columns[0].stats ?? [];

       // find the stat with key="distinct_values"
      const distinctValuesStat = colStats.find(
          (stat) => stat.key === 'distinct_values'
      );
      if (distinctValuesStat?.value instanceof Array) {
          return distinctValuesStat?.value;
      }

      return [];
```

querybook/webapp/hooks/queryEditor/extensions/useAutoCompleteExtension.ts
```diff
@@ -118,13 +95,56 @@ export const useAutoCompleteExtension = ({
...
+            //  - table: schema.table, table
+            //  - keyword: any keyword
+            const generalTokenRegex = /(\w+\.){0,2}\w*/;
+
+            const columnValueToken = context.matchBefore(columnValueRegex);
+            const generalToken =
+                !columnValueToken && context.matchBefore(generalTokenRegex);
+
+            // no token before the cursor, don't open completions.
+            if (!columnValueToken?.text && !generalToken.text) return null;
```
czgu, line 126: generalToken?.text

czgu (approved):

PR #1531, opened 2025-01-03
feat: nlp query search

zhangvi7:
Implement NLP query search to show similar queries. Same as NLP table search
 
image


jczhong84 (commented):
could you also update the `ai_assistant.mdx`?

jczhong84 (approved):

PR #1532, opened 2025-01-13
fix: make sure the cmd-f still work on readonly docs

jczhong84:
Issue: search & replace doesn't show up when the datadoc is readonly
Fix: make the editor readonly but editable (it will respond to keymap events still, but content can not be changed)


zhangvi7 (approved):
thanks for quick fix!

Issue #1484, opened 2024-08-06
Alternating row colors for results


wlfbck:
The reason why should be self explanatory.

Would be really sweet to have this, maybe as a setting. Would make handling wider tables much easier.

PR #1545, opened 2025-03-10
Added default alternating colors in the table rows 

sjay8:
Added default alternating colors used existing defining colors following odd- default, even- darker scheme
#1484


sjay8:
Would it be better to have this as an optional setting rather that the default? Also, would it be better to define a new color specifically for the alternating rows as currently I'm using the color defined for the border? Thank you for your clarifications!

jczhong84:
@sjay8 thanks for putting the PR! could you add it as a user setting? I wonder users may have different preferences.

ronodhirSoumik:
Just to know- dont we need to consider these combinations are okay for all different user themes (default, dark or lush mode)?

PR #1548, opened 2025-03-14
chore: update to use PAT for dockerhub login

jczhong84:
Dockerhub now needs to use PAT for login
> Error response from daemon: Get "https://registry-1.docker.io/v2/": unauthorized: your account must log in with a Personal Access Token (PAT) - learn more at [docs.docker.com/go/access-tokens](http://docs.docker.com/go/access-tokens)


czgu (approved):

jczhong84:
still use the password  key

PR #1550, opened 2025-03-18
feat: add Pyodide Python kernel

jczhong84:
Adding a Python kernel using Pyodide, consisting of the following files:

- python-worker.ts: A web worker that operates with Pyodide.
- python-provider.tsx: A React context provider designed to initialize the Python web worker.
- usePython.ts: A custom hook to interface with the provider context and handle the standard output (stdout).


querybook/webapp/lib/python/patch.ts
```diff
@@ -0,0 +1,115 @@
...
+              df = arg.reset_index()
+              columns = df.columns.tolist()
+              records = df.to_dict(orient='records')
+              _original_print(json.dumps({"type":"dataframe", "columns":columns, "records":records}))
+
+          elif isinstance(arg, (list, dict)):
+              try:
+                  # Try to serialize and print the data as JSON
+                  _original_print(json.dumps({"type":"json", "data": arg}))
+              except:
```
czgu, line 44: except TypeError:

querybook/webapp/lib/python/python-provider.tsx
```diff
@@ -0,0 +1,170 @@
...
+                await initKernel();
+            }
+
+            if (!kernelRef.current) {
+                stderrCallback('Failed to initialize Python kernel');
+            }
+
+            if (status !== PythonKernelStatus.IDLE) {
+                stderrCallback('Kernel is not ready to run code');
+            }
```
czgu, line 132: are we not allowing users running multiple python cells?

querybook/webapp/lib/python/python-worker.ts
```diff
@@ -0,0 +1,219 @@
...
+
+        await this.loadPyodidePromise;
+
+        if (!this.pyodide || !this.interruptBuffer) {
+            throw new Error('Pyodide initialization failed');
+        }
+
+        // Reset interrupt buffer and set status
+        this.interruptBuffer[0] = 0;
+        this.executionCountByNS[namespaceId] =
```
czgu, line 65: namespaceid could be undefined, which is kind of confusing to set in a map, could we put some default name space then?

querybook/webapp/lib/python/python-worker.ts
```diff
@@ -0,0 +1,219 @@
...
+            this._customPrint(result);
+        }
+    }
+
+    /**
+     * Cancel the current Python execution
+     */
+    public cancelRun(): void {
+        if (this.interruptBuffer) {
+            this.interruptBuffer[0] = 2;
```
czgu, line 98: could you put ENUM on top of the values to make it more clear?

czgu (commented):
could you merge it to a feature branch for now?

querybook/webapp/lib/python/python-worker.ts
jczhong84: added undefined check. will create a temporary namespace for any adhoc run without namespace id. dont want to persist any variables using a default namespace.

querybook/webapp/lib/python/python-provider.tsx
jczhong84: yes, we allow.

czgu: i meant what happens if you execute cell 2 while cell 1 is running? it doesn't seem like there is a wait queue at the moment?

czgu (approved):

PR #1553, opened 2025-03-21
feat: add a task queue and support cancelling

jczhong84:
* add a simple queue to run the tasks in sequence. User can kick off multiple runs, and they will become pending and pushed to the queue
* add cancellation support by using shared array buffer, which requires [cross-origin isolated](https://developer.mozilla.org/en-US/docs/Web/API/Window/crossOriginIsolated) enabled.
* add a `progressCallback` to better track the execution status and execution count.
* add another helper function to get identifiers of a namespace to support auto completion.

### DIFF
--- a/querybook/webapp/lib/python/task-queue.ts
+++ b/querybook/webapp/lib/python/task-queue.ts
@@ -1,0 +1,42 @@
+export interface Task {
+    run: () => Promise<void>;
+    resolve: () => void;
+    reject: (error: any) => void;
+}
+
+/**
+ * A simple task queue that runs tasks in sequence.
+ */
+export class TaskQueue {
+    private queue: Task[] = [];
+    private isRunning: boolean = false;
+
+    // Add a task to the queue and immediately attempt to run the next task
+    public push(task: Task) {
+        this.queue.push(task);
+        this.runNext();
+    }
+
+    private async runNext() {
+        // If a task is already running or queue is empty, do nothing.
+        if (this.isRunning || this.queue.length === 0) {
+            return;
+        }
+
+        // Pull the next task from the queue.
+        const task = this.queue.shift();
+        if (!task) return;
+
+        this.isRunning = true;
+        try {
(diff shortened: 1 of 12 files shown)


querybook/webapp/lib/python/usePython.ts
```diff
@@ -29,10 +36,21 @@ export default function usePython({
 }: UsePythonProps): UsePythonReturn {
     const [stdout, setStdout] = useState<string[]>([]);
     const [stderr, setStderr] = useState<string[]>([]);
+    const [executionStatus, setExecutionStatus] =
+        useState<PythonExecutionStatus>();
```
czgu, line 40: can you assign a default executionStatus? unless undefined meant for anything

querybook/webapp/lib/python/usePython.ts
```diff
@@ -29,10 +36,21 @@ export default function usePython({
...
+    const [executionCount, setExecutionCount] = useState<number>();
 
-    const { status, getExecutionCount, runPython, createDataFrame } =
+    const { status, runPython, cancelRun, createDataFrame, getNamespaceInfo } =
         useContext(PythonContext);
 
+    const progressCallback = useCallback(
```
czgu, line 46: can use the same typing signature here

querybook/webapp/lib/python/task-queue.ts
```diff
@@ -0,0 +1,42 @@
...
+        this.isRunning = true;
+        try {
+            await task.run();
+            task.resolve();
+        } catch (error) {
+            task.reject(error);
+        } finally {
+            this.isRunning = false;
+            // Process the next task in queue.
+            this.runNext();
```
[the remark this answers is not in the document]
jczhong84: I think we can still keep running the remaining tasks? for those cells depend on the error cell, they will also error out. and for others, they can still run successfully.

czgu (approved):

PR #1554, opened 2025-03-21
feat: add python editor

jczhong84:
Add editor for the Python cell.

Extracted the common code mirror part as a new component `CodeEditor` and both `QueryEditor` and `PythonEditor` will use it.

Snip20250321_10
Snip20250321_12


querybook/webapp/components/CodeEditor/CodeEditor.tsx
```diff
@@ -0,0 +1,182 @@
...
+            [onSelection]
+        );
+
+        const searchExtension = useSearchExtension({
+            editorView: editorRef.current?.view,
+            cellId,
+        });
+
+        const onChangeHandler = useCallback(
+            (value, viewUpdate) => {
```
czgu, line 122: value: string

is viewUpdate unused?

czgu (approved):

querybook/webapp/components/CodeEditor/CodeEditor.tsx
jczhong84: yes, it's unused.

PR #1555, opened 2025-04-02
feat: add python cell

jczhong84:
changes
 - Add a new datadoc cell type: python cell with title and python editor
 - Update db model for the python cell result
 - Add python cell result view to support json, dataframe, image and plain text
 - Add a "Copy DataFrame" button for the statement result
 - An update to the `get_df` function to prevent getting statement result from a different environment which doesn't support that query engine.

Snip20250402_25
Snip20250402_26
Snip20250402_30
Snip20250402_29

`get_df`
image
image


querybook/webapp/components/DataDocPythonCell/DataDocPythonCell.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const fetchNamespaceIdentifiers = useCallback(async () => {
+        const info = await getNamespaceInfo?.(docId);
+        if (info) {
+            setIdentifiers(info.identifiers);
+        }
+    }, [docId, getNamespaceInfo, setIdentifiers]);
+
+    const runPythonCode = useCallback(async () => {
+        const result = await runPython(context);
+        console.log(result);
```
czgu, line 71: remove?

querybook/webapp/components/DataDocPythonCell/PythonCellResultView.tsx
```diff
@@ -0,0 +1,76 @@
...
+        const records = data['records'];
+
+        return [
+            columns,
+            ...records.map((record) => columns.map((col) => record[col])),
+        ];
+    }, []);
+
+    const renderOutput = (output) => {
+        if (typeof output === 'object') {
```
czgu, line 36: what if it is output object without a type key?

querybook/webapp/components/DataDocPythonCell/PythonCellResultView.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { vscodeTheme } from '@uiw/react-json-view/vscode';
+import React, { useCallback, useMemo } from 'react';
+
+import { StatementResultTable } from 'components/StatementResultTable/StatementResultTable';
+import { useShallowSelector } from 'hooks/redux/useShallowSelector';
+import { IStoreState } from 'redux/store/types';
+
+import './PythonCellResultView.scss';
+
+export const PythonCellResultView = ({ stdout, stderr }) => {
```
czgu, line 12: prop types are missing

querybook/webapp/lib/python/usePython.ts
```diff
@@ -9,44 +11,78 @@ import {
...
+    useEffect(() => {
+        if (
+            cellId &&
+            (executionStatus === PythonExecutionStatus.SUCCESS ||
+                executionStatus === PythonExecutionStatus.ERROR ||
+                executionStatus === PythonExecutionStatus.CANCEL)
+        ) {
+            PythonCellResource.updateResult(cellId, stdout, stderr);
+        }
+    }, [executionStatus, stdout, stderr]);
```
czgu, line 72: this useEffect already triggers on stdout/stderr changes, are we basically make sure no intermediate states are saved?

querybook/webapp/lib/python/usePython.ts
```diff
@@ -9,44 +11,78 @@ import {
...
     onStdout,
     onStderr,
 }: UsePythonProps): UsePythonReturn {
     const [stdout, setStdout] = useState<string[]>([]);
-    const [stderr, setStderr] = useState<string[]>([]);
+    const [stderr, setStderr] = useState<string>();
     const [executionStatus, setExecutionStatus] =
         useState<PythonExecutionStatus>();
     const [executionCount, setExecutionCount] = useState<number>();
+    const cancelPromiseResolveRef = useRef(null);
```
czgu, line 46: remember to add typing

querybook/webapp/lib/python/usePython.ts
```diff
@@ -108,10 +144,17 @@ export default function usePython({
         [setStdout, setStdout, runPython, docId, stdoutCallback, stderrCallback]
     );
 
+    const cancelRunPython = useCallback(async (): Promise<void> => {
```
czgu, line 147: what's the point of async?

czgu (commented):
can you add typing and validation to python for the stdout that's stored?

querybook/webapp/components/DataDocPythonCell/PythonCellResultView.tsx
jczhong84: it will fallback to the plain text view

querybook/webapp/lib/python/usePython.ts
jczhong84: to only trigger the result save to db only happens once (after the cell execution finishes). stdout may change multiple times for the cell run.

querybook/webapp/lib/python/usePython.ts
jczhong84: cancel will not actually happen immediately after setting the signal buffer. This is to simulate it to be a promise and resolve after the cancel event actually happens.

querybook/webapp/lib/python/usePython.ts
czgu: hmm, yeah I was wondering if we should just do a debounced save instead

querybook/webapp/lib/python/usePython.ts
czgu: there is no await here so I dont think the async does anything?

jczhong84: oh yeah, correct. the "aync" is not needed.

querybook/server/datasources/python_cell.py
```diff
@@ -1,25 +1,71 @@
...
 
 
+class PythonOutputDataFrameDataType(TypedDict):
+    columns: List[str]
+    records: List[Dict[str, Any]]
+
+
+class PythonOutputDictType(TypedDict):
+    type: Literal["dataframe", "image", "json"]
+    data: Union[str, PythonOutputDataFrameDataType]
```
czgu, line 18: nit: can you move this and the validation logic to another file?

querybook/webapp/lib/python/types.ts
```diff
@@ -51,6 +51,24 @@ export enum InterruptBufferStatus {
     SIGINT = 2, // Interrupt the execution
 }
 
+export type PythonOutputType =
```
czgu, line 54: have some comments to make sure backend frontend typing are consistent

querybook/webapp/lib/python/types.ts
```diff
@@ -51,6 +51,24 @@ export enum InterruptBufferStatus {
...
+    | {
+          type: 'dataframe';
+          data: {
+              columns: string[];
+              records: Record<string, any>[];
```
czgu, line 59:
```suggestion
              records: Array<Record<string, any>>;
```

czgu (approved):

PR #1561, opened 2025-04-10
fix: get task run records by exact name when disabling datadoc

xuantu2022:
Fuzzy matching by name caused successes of another datadoc to be incorrectly included, which prevented a continuously failing datadoc from being disabled.

image

Use `get_task_run_record_run_by_name` instead to perform exact name matching and resolve this isssue.


jczhong84 (approved):
thanks for the fix!

jczhong84 (approved):

PR #1562, opened 2025-04-14
fix: template query hangs with big payload

jczhong84:
**Summary**
This PR addresses a vulnerability in the /ds/query_execution/templated_query/ endpoint of Querybook, where a Regular Expression used to escape SQL comments is vulnerable to ReDoS attacks.

**Impact**
The flaw allows an attacker to send malicious queries that can hang the server and consume excessive CPU resources, impacting availability.

**Reproduction Steps**
Attacker logs in and sends a crafted POST request to the endpoint.
The server hangs and CPU usage spikes to nearly 100%.
See the new test case as an example

**Root Cause**
The issue stems from the use of re.sub with a vulnerable regex pattern in the _escape_sql_comments function.

**Proposed Fix**
Replace the regex-based substitution with a non-regex alternative.

### DIFF
--- a/querybook/server/lib/query_analysis/templating.py
+++ b/querybook/server/lib/query_analysis/templating.py
@@ -49,19 +49,19 @@ def _escape_sql_comments(query: str):
     query_len = len(query)
     while pos < query_len:
         ch = query[pos]
-        # If inside a quoted string, just copy it verbatim.
-        if ch in ("'", '"'):
-            quote_char = ch
+        # If inside a quoted string, just copy it verbatim. ANSI SQL only allows single quotes.
+        if ch == "'":
             start = pos
             pos += 1
             while pos < query_len:
-                if query[pos] == "\\":
-                    pos += 2
-                    continue
-                if query[pos] == quote_char:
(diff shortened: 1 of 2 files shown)


czgu (approved):

querybook/server/lib/query_analysis/templating.py
```diff
@@ -34,15 +33,58 @@ class LatestPartitionException(QueryTemplatingError):
...
+    while pos < n:
+        ch = query[pos]
+        # If inside a quoted string, just copy it verbatim.
+        if ch in ("'", '"'):
+            quote_char = ch
+            start = pos
+            pos += 1
+            while pos < n:
+                if query[pos] == "\\":
+                    pos += 2
```
czgu, line 49: add a comment here explaining we are skipping 2 chars because its \' or \"

jczhong84: actually ANSI only uses single quote and uses doubled single quote to escape the single quote. will update

czgu (approved):

PR #1566, opened 2025-04-19
fix: kernel status is incorrect with mutlipel task running

jczhong84:
The issue: the kernel status will be cleared as idle when there is still pending tasks.

Fix: switch to derive the busy and idle status from the python execution status.


czgu (commented):
i dont think it makes sense to expose setKernelStatus to consumers as it just makes the logic more scattered in different places. Have you tried the having a task counter approach I mentioned?

jczhong84:
> i dont think it makes sense to expose setKernelStatus to consumers as it just makes the logic more scattered in different places. Have you tried the having a task counter approach I mentioned?

you're right! updated to use a counter.

czgu (approved):

PR #1576, opened 2025-05-02
refactor: allow ai assistant plugin to override sql gen and fix

jczhong84:
- add two functions `_generate_sql_query` and `_query_auto_fix` in `base_ai_assistant` for plugin to override easlily
 - upgrade new langchain packages
 - add openai new models

### DIFF
--- a/querybook/server/lib/ai_assistant/assistants/openai_assistant.py
+++ b/querybook/server/lib/ai_assistant/assistants/openai_assistant.py
@@ -17,8 +17,11 @@ OPENAI_MODEL_CONTEXT_WINDOW_SIZE = {
     # Current models
     "gpt-4o": 128000,
     "gpt-4o-mini": 128000,
+    "gpt-4.1": 1_047_576,
+    "gpt-4.1-mini": 1_047_576,
+    "gpt-4.1-nano": 1_047_576,
 }
-DEFAULT_MODEL_NAME = "gpt-4o-mini"
+DEFAULT_MODEL_NAME = "gpt-4.1-mini"
 
 
 class OpenAIAssistant(BaseAIAssistant):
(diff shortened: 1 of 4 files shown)


querybook/server/lib/ai_assistant/assistants/openai_assistant.py
```diff
@@ -45,7 +48,11 @@ def _get_default_llm_config(self):
 
     def _get_token_count(self, ai_command: str, prompt: str) -> int:
         model_name = self._get_llm_config(ai_command)["model_name"]
-        encoding = tiktoken.encoding_for_model(model_name)
+        if model_name.startswith("gpt-4.1"):
```
czgu, line 51: better to add a comment why, OpenAI needs to update Tiktoken library!

querybook/server/lib/ai_assistant/base_ai_assistant.py
```diff
@@ -314,10 +331,29 @@ def query_auto_fix(
...
+            language=language,
+            query=query,
+            error=self._get_query_execution_error(query_execution),
+            session=session,
+        )
+
+    @with_session
+    def _query_auto_fix(
+        self,
+        socket,
```
czgu, line 346: position of this socket is kind of all over the space, consider moving it to be before session

querybook/server/lib/ai_assistant/base_ai_assistant.py
```diff
@@ -254,8 +254,25 @@ def generate_sql_query(
...
+            socket=socket,
+        )
+
+    def _generate_sql_query(
+        self,
+        language: str,
+        question: str,
+        table_schemas: list[str],
+        original_query: str = None,
+        socket=None,
```
czgu, line 271: is socket not required here? how come the autofix one requires it

querybook/webapp/components/AIAssistant/AutoFixButton.tsx
```diff
@@ -31,6 +33,7 @@ const useSQLFix = () => {
         data: unformattedData,
         explanation,
         fix_suggestion: suggestion,
+        diagnosis,
```
czgu, line 36: how did we end with a new field without backend change?

kgopal492 (approved):

querybook/server/lib/ai_assistant/base_ai_assistant.py
jczhong84: yes, it is required actually. let me remove the default value.

czgu (approved):

PR #1580, opened 2025-05-12
chore: Optimize SQL performance for tables with large number of columns

jczhong84:
This PR introduces performance optimizations for tables with many columns by reducing database load and query count:

### Key Changes:
- **Batch loading**: Implemented consolidated batch queries for columns, tags, and data element associations instead of individual queries
- **Optimized `get_table_schemas_by_names`**: Completely rewritten to:
  - Fetch tables, columns, and tags in separate efficient queries
  - Use SQLAlchemy's `load_only` and `joinedload` to minimize data transfer
  - Process results in memory rather than making many small queries
- **Optimized `get_detailed_columns_dict_by_table_id`**: Reimplemented to:
  - Retrieve all column data in a single query
  - Batch load tags and data element associations for all columns at once
  - Assemble column data in memory instead of making separate queries per column
- **Model improvements**:
  - Added `lazy="joined"` optimization for column statistics as `stats` will be retrieved by default
  - Added `to_dict()` method for DataTableColumnStatistics to support JSON serialization, which will make `flask.jsonify` more efficient.

### Performance Improvements:
- **Single table with 581 columns**:
  - `/ds/table/<table-id>/`: ~10s → <1s
  - `/ds/table/<table-id>/detailed_column/`: 5-6s → <1s
- **Multiple tables (10 tables including large table)**:
  - Schema retrieval: 9-10s → <2s

### DIFF
--- a/querybook/server/lib/ai_assistant/tools/table_schema.py
+++ b/querybook/server/lib/ai_assistant/tools/table_schema.py
@@ -1,23 +1,16 @@
 from typing import Callable
 
+
+from sqlalchemy import and_, or_
+from sqlalchemy.orm import joinedload, load_only
 from app.db import with_session
-from lib.vector_store import get_vector_store
 from logic import metastore as m_logic
-from models.metastore import DataTable, DataTableColumn
-
-
-def get_table_documentation(table: DataTable) -> str:
-    """Get table documentation.
-    Try to get it from database first. If not found, get it from the vector store.
-    """
-    if table.information.description:
-        return table.information.description
-
-    vs = get_vector_store()
-    if vs:
-        return vs.get_table_summary(table.id)
-
-    return ""
+from models.metastore import (
+    DataSchema,
+    DataTable,
+    DataTableColumn,
+)
+from models.tag import TagItem
 
 
 def _get_column(column: DataTableColumn) -> dict[str, str]:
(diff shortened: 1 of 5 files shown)


querybook/server/logic/data_element.py
```diff
@@ -77,6 +77,53 @@ def get_data_element_association_by_column_id(
     return data_element
 
 
+@with_session
+def get_data_element_associations_by_column_ids(column_ids: list[int], session=None):
```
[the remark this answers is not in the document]
czgu: should this just be `get_column_to_data_element_mapping`

querybook/server/logic/tag.py
```diff
@@ -29,6 +29,31 @@ def get_tags_by_column_id(column_id: int, session=None):
     )
 
 
+@with_session
+def get_tags_by_column_ids(column_ids: int, session=None):
```
czgu, line 33: wrong typing?

czgu: also add a return type

querybook/server/lib/ai_assistant/tools/table_schema.py
```diff
@@ -255,7 +273,7 @@ def get_table_schema_by_name(
 
 def get_slimmed_table_schemas(
     table_schemas: list[dict], column_keys_to_keep=["name", "type"]
-) -> list[dict]:
+) -> list[TableSchema]:
```
czgu, line 276: typing doesn't look right, also can you update the caller typing for table_schemas

czgu (approved):

PR #1679, opened 2026-03-26
Add stale query warning with revert button for query executions

wbh123456:
#### Add execution snapshots and stale query warning
When a user edits a query after running it, the displayed execution results may no longer match the current query text. This PR adds a "stale query" warning banner that appears when the editor content diverges from what was actually executed. The warning includes a Revert button that restores the editor to the exact query that produced the selected execution.

**Key changes:**
- useExecutionSnapshots hook to capture the query text at execution time
- useStaleQueryWarning hook with debounced diff detection
- StaleQueryWarning component with revert action
- Integrated into both DataDoc query cells and the Query Composer
- Unit tests for snapshot storage and stale detection logic

#### Minor Changes
- Consolidate the content on AGENTS.md and make the CLAUDE.md to point to a single source.

## Jira
https://pinterest.atlassian.net/browse/AP-5669

## Demo

https://github.com/user-attachments/assets/252d4b38-5c87-4ada-9cb4-0fc361a7af9f

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -56,16 +56,29 @@ make scheduler  # celery beat
 
 ## Making Commits
 
-When preparing a PR, run the relevant checks:
+When preparing a PR, run the relevant checks. CI runs all of the following automatically on every PR via GitHub Actions (`.github/workflows/`).
 
 **Backend changes** (anything under `querybook/server/`):
-- `make test` — run backend tests
-- Pre-commit hooks handle linting (configured in `.pre-commit-config.yaml`)
+- `make test` — builds the `querybook-test` Docker image and runs pytest inside it
+- CI-equivalent without Docker: `PYTHONPATH=querybook/server:plugins ./querybook/scripts/run_test --python`
 
 **Frontend changes** (anything under `querybook/webapp/`):
-- `pnpm test` or `yarn test` — run Jest tests
-- `pnpm run lint` or `yarn lint` — ESLint with auto-fix
-- `pnpm run tsc-check` or `yarn tsc-check` — TypeScript type checking
+- `./querybook/scripts/run_test --node` — CI-equivalent; runs all four checks below in parallel:
+  - `yarn tsc-check` — TypeScript type checking
+  - `yarn test` — Jest unit tests
+  - `yarn lint` — ESLint with auto-fix
+  - `webpack --mode=production` — production build verification
+
+**Formatting (all changes) — common CI failure:**
+
+`./querybook/scripts/run_test --node` does **not** run Prettier. CI runs Prettier separately via `pre-commit`, so formatting issues are a frequent cause of CI failures. Always run Prettier on changed files before pushing:
+
+- `yarn prettier --write <files>` — format specific files
+- `pre-commit run --all-files` — run the full pre-commit suite (Black, Prettier, flake8)
+
+To catch formatting automatically on every `git commit`, install the hooks once:
+
+- `pip install pre-commit && pre-commit install`
 
 ## Pinterest Internal Deployment
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,91 @@
+# Querybook
+
+Querybook is Pinterest's open-source Big Data IDE for discovering, creating, and sharing data analyses. It combines a rich-text editor, SQL query engine, charting, scheduling, and table documentation in a single web app.
+
+## Tech Stack
+
+- **Backend:** Python 3.10, Flask, SQLAlchemy (MySQL), Celery (Redis broker), Elasticsearch/OpenSearch, gevent + Flask-SocketIO (WebSockets), uWSGI (production)
+- **Frontend:** React 17, TypeScript, Redux, Webpack 5, CodeMirror (SQL editor), Draft.js (rich text), Chart.js/D3/ReactFlow
+
+## Directory Layout
+
+- `querybook/server/` — Flask backend
+  - `app/` — app setup
+  - `datasources/` — REST API endpoints
+  - `logic/` — business logic
+  - `models/` — SQLAlchemy models
+  - `tasks/` — Celery tasks
+  - `lib/` — utilities, executors, metastores
+  - `env.py` — `QuerybookSettings` configuration
+- `querybook/webapp/` — React/TypeScript frontend
+  - `components/` — React components
+  - `hooks/` — custom React hooks
+  - `redux/` — Redux store, actions, reducers
+  - `lib/` — frontend utilities
+  - `ui/` — reusable UI primitives
+  - `resource/` — API client layer
+- `querybook/config/` — YAML config files
+- `plugins/` — plugin stubs (extension point for custom behavior)
+- `requirements/` — pip requirements (`base.txt`, `prod.txt`, `engine/*.txt`, `auth/*.txt`)
+- `containers/` — Docker Compose files (dev, prod, test)
+- `docs_website/` — Docusaurus documentation site
+- `helm/` / `k8s/` — Kubernetes deployment manifests
+
+## Plugin System
+
+Querybook is extended via plugins without forking. The env var `QUERYBOOK_PLUGIN` (default `./plugins`) points to a directory where plugin modules are discovered by `lib.utils.import_helper.import_module_with_default()`.
+
+Each plugin module exports a well-known variable (e.g. `ALL_PLUGIN_EXECUTORS`) that the server merges with built-in defaults.
+
+Key plugin types: `executor_plugin`, `metastore_plugin`, `auth_plugin`, `api_plugin`, `exporter_plugin`, `result_store_plugin`, `notifier_plugin`, `event_logger_plugin`, `stats_logger_plugin`, `job_plugin`, `tasks_plugin`, `dag_exporter_plugin`, `ai_assistant_plugin`, `vector_store_plugin`, `webpage_plugin`, `monkey_patch_plugin`, `query_validation_plugin`, `query_transpilation_plugin`, `engine_status_checker_plugin`, `table_uploader_plugin`.
+
+## Configuration
+
+Priority: **env vars > `querybook_config.yaml` > `querybook_default_config.yaml`**.
+
+Key settings live in `querybook/server/env.py` (`QuerybookSettings`).
+
+## Running Locally
+
+```bash
+make            # docker-compose up (full stack) → http://localhost:10001
+make web        # web server only
+make worker     # celery worker
+make scheduler  # celery beat
+```
+
+## Making Commits
+
+When preparing a PR, run the relevant checks:
+
+**Backend changes** (anything under `querybook/server/`):
+- `make test` — run backend tests
+- Pre-commit hooks handle linting (configured in `.pre-commit-config.yaml`)
+
+**Frontend changes** (anything under `querybook/webapp/`):
+- `pnpm test` or `yarn test` — run Jest tests
+- `pnpm run lint` or `yarn lint` — ESLint with auto-fix
+- `pnpm run tsc-check` or `yarn tsc-check` — TypeScript type checking
+
+## Pinterest Internal Deployment
+
+Pinterest deploys Querybook internally via a separate repo (`datahub-pinterest`) that uses this repo's Docker image as a base and layers Pinterest-specific plugins on top. Changes to core Querybook go here; Pinterest-specific features belong in that plugin repo. Do not add Pinterest-internal details to this file.
+
+## Maintaining This File
+
+**Include:**
+- Repo purpose, tech stack, and high-level architecture
+- Directory layout (key paths only)
+- How to run, test, and lint locally
+- Commit and PR workflow expectations
+- Plugin system overview and extension points
+
+**Do not include:**
+- Detailed API docs or function-level documentation
+- Inline code examples longer than 5 lines
+- Deployment runbooks or operational procedures (keep in README or docs/)
+- Credentials, secrets, or internal URLs
+- Information that changes frequently (version numbers, dependency lists)
+- Content already covered in README.md
+- Content that can be easily derived by AI agents (e.g. reading file trees, package.json)
+- References to internal/proprietary repos — this is an open-source project
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,8 +47,17 @@ Key settings live in `querybook/server/env.py` (`QuerybookSettings`).
 
 ## Running Locally
 
+Start the full stack (web server, worker, scheduler, and all dependencies) with Docker Compose:
+
 ```bash
-make            # docker-compose up (full stack) → http://localhost:10001
+make
+```
+
+This brings up everything and serves the app at http://localhost:10001. This is the primary command for local development.
+
+To restart individual services without bouncing the full stack:
+
+```bash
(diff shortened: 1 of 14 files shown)


docs_website/docs/developer_guide/contributing.mdx
```diff
@@ -65,6 +65,41 @@ To increase the chances that your pull request will be accepted:
 -   Write tests for your changes
 -   Write a good commit message
 
+### Testing
+
+CI runs all of the checks below automatically on every pull request via GitHub Actions. You should reproduce them locally before pushing.
```
jasoncoffman: Was this recently changed? AFAIK if the person who opened the PR isn't a maintainer, it needs to be manually kicked off by a maintainer

docs_website/docs/developer_guide/contributing.mdx
```diff
@@ -65,6 +65,41 @@ To increase the chances that your pull request will be accepted:
...
+
+```sh
+# CI-equivalent — runs TypeScript check, Jest, ESLint, and a production webpack build in parallel
+./querybook/scripts/run_test --node
+
+# Or run individual checks manually
+yarn tsc-check   # TypeScript type checking
+yarn test         # Jest unit tests
+yarn lint         # ESLint with auto-fix
+```
```
jasoncoffman: `make test` runs both the python and node tests by default, can we just recommend one command for testing

querybook/webapp/components/DataDocQueryExecutions/StaleQueryWarning.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const StaleQueryWarning: React.FC<IProps> = ({ onRevert }) => (
+    <Message
+        type="warning"
+        icon="AlertTriangle"
+        iconSize={16}
+        size="small"
+        message={
+            <span className="flex-row" style={{ gap: 8 }}>
+                Query has been edited.
+                {onRevert && (
```
jasoncoffman: I'm not sure we want a revert action here - it also may be a bit too easy to click and i would guess that the vast majority of the time the changes are intentional; `cmd+z` works great for a local undo.

If we want a revert I think adding it in the executions drop down would be a better place. That way users can view each execution of a cell and have the option to effectively rollback to the previous execution. But that should likely be it's own separate feature

querybook/webapp/components/DataDocQueryExecutions/StaleQueryWarning.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import { TextButton } from 'ui/Button/Button';
+import { Message } from 'ui/Message/Message';
+
+interface IProps {
+    onRevert?: () => void;
+}
+
+export const StaleQueryWarning: React.FC<IProps> = ({ onRevert }) => (
+    <Message
```
jasoncoffman: Can we make the warning a bit less intrusive; I imagine the majority of users are intentionally making change.

Can we hook into the execution state indicator below the cell and add a new state (with tooltip) there? This is more inline with the patterns that other notebook tools use

AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+- `pre-commit run --all-files` — run the full pre-commit suite (Black, Prettier, flake8)
+
+To catch formatting automatically on every `git commit`, install the hooks once:
+
+- `pip install pre-commit && pre-commit install`
+
+## Pinterest Internal Deployment
+
+Pinterest deploys Querybook internally via a separate repo (`datahub-pinterest`) that uses this repo's Docker image as a base and layers Pinterest-specific plugins on top. Changes to core Querybook go here; Pinterest-specific features belong in that plugin repo. Do not add Pinterest-internal details to this file.
+
```
jasoncoffman: Let's not include this in the oss repo

AGENTS.md
```diff
@@ -0,0 +1,104 @@
+# Querybook
+
+Querybook is Pinterest's open-source Big Data IDE for discovering, creating, and sharing data analyses. It combines a rich-text editor, SQL query engine, charting, scheduling, and table documentation in a single web app.
+
+## Tech Stack
+
+- **Backend:** Python 3.10, Flask, SQLAlchemy (MySQL), Celery (Redis broker), Elasticsearch/OpenSearch, gevent + Flask-SocketIO (WebSockets), uWSGI (production)
+- **Frontend:** React 17, TypeScript, Redux, Webpack 5, CodeMirror (SQL editor), Draft.js (rich text), Chart.js/D3/ReactFlow
```
jasoncoffman, line 8: we should include pinned versions for each of these if we are going to mention them here otherwise we might 
a) generate code that doesn't work with the versions we are using
b) end up loading `requirements.txt` or `package.json` to get version info

AGENTS.md
jasoncoffman, line 1: This is probably fine for now, but if this starts growing we should follow progressive disclosure principles

AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Directory Layout
+
+- `querybook/server/` — Flask backend
+  - `app/` — app setup
```
jasoncoffman, line 13: can we either remove or improve this - there is a lot of content in `app` (auth clients for all auth providers, datasource decorators, database client)

AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+  - `datasources/` — REST API endpoints
```
jasoncoffman, line 14: any reason to leave out the `client`, `datasource_socketio`, `scheduler`, `scripts` directories

AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+Key settings live in `querybook/server/env.py` (`QuerybookSettings`).
+
+## Running Locally
+
+```bash
+make            # docker-compose up (full stack) → http://localhost:10001
+make web        # web server only
+make worker     # celery worker
+make scheduler  # celery beat
+```
```
jasoncoffman, line 64: This might cause issues having there all in one code block with minimal comments. We should be more explicit (both in instruction and separating out the commands) that `make` is the primary method for starting all containers for development, and each of those other targets for for starting up individual services

AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Making Commits
+
+When preparing a PR, run the relevant checks. CI runs all of the following automatically on every PR via GitHub Actions (`.github/workflows/`).
```
jasoncoffman: this was an issue before I was a registered maintainer. These must be manually triggered by a maintainer

AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+
+**Backend changes** (anything under `querybook/server/`):
+- `make test` — builds the `querybook-test` Docker image and runs pytest inside it
```
jasoncoffman: For the agents we can probably just instruct to only use `make test` so that it's an isolated, reproducible environment that the tests are run in

docs_website/docs/developer_guide/contributing.mdx
wbh123456: Nice catch, I changed the wording to reflect this.

querybook/webapp/components/DataDocQueryExecutions/StaleQueryWarning.tsx
wbh123456: Right, this is also something I hesitated to add. 

Since it is not difficult to add, I think I can just remove the revert button for now to keep it separated from the concern. We can always add it back in the future if needed.

AGENTS.md
wbh123456: This could be helpful to AI agents to cross referencing the codes if they are both in the local. The section here only specifies a minimal of info with the repo name. I believe it is not too sensitive to share this info in this repo.

WDYT on this?

querybook/webapp/components/DataDocQueryExecutions/StaleQueryWarning.tsx
wbh123456: I agree with you that it is a bit intrusive. I made it into a banner on purpose because we want to warn the user the query has been edited without an execution to remind them to perform the execution. This will also avoid the confusion of sending AI the staled query.

While making the warning too large makes it too intrusive, making it too small defeats the purpose of the warning.

I will play around with a few different UX and let's sync offline on this UI design and find a middle ground.

AGENTS.md
wbh123456: These are left out on purpose (by AI)
Since the AGENTS.md is meant to be read by AI and passed into context, I asked the AI to only leave key paths here that could be beneficial for the agents. Therefore, it decided what were some key paths worth noting down vs. what could be easily be derived simply at run time on its own.

Check ## Maintaining This File section (Line 87) in the same file as a reference.

AGENTS.md
wbh123456: These are left out on purpose (by AI)
Since the AGENTS.md is meant to be read by AI and passed into context, I asked the AI to only leave key paths here that could be beneficial for the agents. Therefore, it decided what were some key paths worth noting down vs. what could be easily be derived simply at run time on its own.

Check ## Maintaining This File section (Line 87) in the same file as a reference.

AGENTS.md
wbh123456: Change the wording to reflect this.

AGENTS.md
jasoncoffman: Definitely agree that this isn't too sensitive! My concern is more that this is going to pollute the context of any users outside of Pinterest that use a coding agent; since this is a top level agents file it'll likely always get picked up and could cause issues

If we have issues with agents not knowing where certain code should live, we should iterate on the datahub agents

querybook/webapp/hooks/queryEditor/useStaleQueryWarning.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export type StaleWarningState = 'edited' | 'unsaved' | null;
+
+export function shouldComputeStaleWarning(
+    selectedExecutionId: number | null | undefined,
+    snapshots: Readonly<Record<number, string>>,
+    currentInput: string,
+    initialQuery?: string
+): boolean {
+    if (selectedExecutionId == null) {
```
jasoncoffman, line 13:
```suggestion
    if (selectedExecutionId === null) {
```

querybook/webapp/hooks/queryEditor/useExecutionSnapshots.ts
jasoncoffman, line 1: Could we just directly access execution state, rather than use snapshotting?

It seems like the current implementation doesn't bootstrap well (if the saved query doesn't match execution on opening a datadoc), and doesn't respect variables changes

querybook/webapp/hooks/queryEditor/useStaleQueryWarning.ts
jasoncoffman, line 1: Does having the separate saved stale check add anything other than allowing us to display different tooltips?  May be worth removing that extra logic to simplify this hook. In the end, we are only trying to indicate the current SQL does not match the viewed execution

querybook/webapp/components/QueryComposer/QueryComposerExecution.tsx
```diff
@@ -55,6 +71,9 @@ export const QueryComposerExecution: React.FunctionComponent<IProps> = ({
                             Execution {execution.id}
                         </span>
                     </AccentText>
+                    {warningState && (
+                        <StaleQueryWarning variant={warningState} />
+                    )}
```
jasoncoffman: Can we render this in the same position as in the datadoc implementation

querybook/webapp/components/QueryComposer/QueryComposer.tsx
```diff
@@ -571,16 +586,17 @@ const QueryComposer: React.FC = () => {
...
                         queryExecutionsAction.createQueryExecution(
-                            query,
+                            q,
                             engineId,
                             null,
                             queryExecutionMetadata,
                             peerReviewParams
                         )
                     );
+                    recordRunInputSnapshot(data.id, query);
```
jasoncoffman, line 599: Any reason we want to snapshot the `query` instead of the query (`q`) that the execution is created with?

querybook/webapp/hooks/queryEditor/useStaleQueryWarning.ts
wbh123456: The loose == null is an intentional to catch both null and undefined

AGENTS.md
wbh123456: > pollute the context of any users outside of Pinterest that use a coding agent
I agree with you. Let me remove this section.

querybook/webapp/hooks/queryEditor/useExecutionSnapshots.ts
wbh123456: This is good point and is it something we have to make tradeoff upon. I think this is something worth discussing:
- **Snapshot approach** (current):  Compares raw editor text at run time vs raw editor text now -- apples-to-apples. But this creates the bootstrapping problem you mentioned above.
- **Execution state approach** (suggested): Always available from Redux on load, no bootstrapping gap. But execution.query is the fully transformed SQL (post-Jinja, post-sampling, post-row-limit). Comparing it against raw editor text would cause false-positive stale warnings on every query using templates or sampling.

I think neither approach is not perfect unless we make some more significant changes into the system to perfectly track the query executions. Not mentioning that there will also be "selected query" vs. "full query in the editor" problem.

I decided to use the **Snapshot approach** approach in the end because it is the most accurate - only with the tradeoff on some false negatives (bootstrapping problem). I believe false negatives are better then false positives in this case. 

So to put in simple words, I dont think there is a perfect solution here but the current approach is the one that works the best. We can discuss more on this offline.

wbh123456:
## Risks with stale-warning badge
### Approach 1: In-Memory Snapshot (in the PR)
Compare the live editor text against a snapshot of the raw editor text captured at run time, stored in React component state.

- Bootstrapping / page refresh (false negative): Snapshots live only in browser memory. After a refresh, the system falls back to the current saved cell content, so edits made before the last run are invisible — no warning appears even though results are stale.
- Cross-session and collaborator edits (false negative): If another user (or the same user in a previous session) edited the cell after the last run, the fallback baseline is the new text, masking the mismatch.
- No server-side raw-input record: QueryExecution.query stores rendered SQL, not what was in the editor. Persisting snapshots server-side would require a schema migration and ongoing storage cost for every execution.

### Approach 2: Compare Against Executed Query in DB
Compare the live editor text against QueryExecution.query (the rendered, actually-executed SQL).

- Templated queries (false positive): The editor contains {{ today }} but the DB stores the resolved date. These never match, so the banner fires permanently on any templated cell.
- Macros (false positive): Inline {% macro %} blocks and macros.ds_add(...) calls expand during rendering. The raw-vs-rendered mismatch guarantees a perpetual warning.
- Multi-statement mismatch (false positive): The comparison is cell-level, but statement splitting, comment stripping, and whitespace normalization mean the stored SQL rarely matches the raw editor text character-for-character, even immediately after running.

### Decision: Not Implementing
While a theoretically perfect solution may exist (e.g., persisting raw input server-side, tracking variable values, doing per-statement diffing), the engineering effort is disproportionate to the benefit. Comparable platforms such as Google Colab and Jupyter do not provide stale-result warnings either. We will leave staleness management to the user/author and remove the prototyped StaleQueryWarning code.

PR #1707, opened 2026-05-14
Create EXTRA_CLIENT_SETTINGS_HOOK to inject flask params into query engine client_settings

kgopal492:
### Context
Today, plugin-provided executor clients receive their kwargs from a single `client_setting` dict that's computed from the `QueryEngine` row in the database. There's no clean way to inject values that are only known at request time (i.e. from the Flask request that triggered the query). This PR adds an optional `EXTRA_CLIENT_SETTINGS_HOOK` plugin hook that lets a plugin compute a dict of extra `client_setting` kwargs once in the originating Flask request and have them flow through every executor entry point — sync, async, and peer-review approval — without any DB plumbing.

### Summary of changes
- **[plugins/executor_plugin/__init__.py](plugins/executor_plugin/__init__.py)** — declares the new optional `EXTRA_CLIENT_SETTINGS_HOOK` symbol (defaults to `None`) with a docstring example showing how a plugin can read the Flask request and return a `Dict[str, Any]` to merge into `client_setting`.
- **[querybook/server/lib/query_executor/executor_factory.py](querybook/server/lib/query_executor/executor_factory.py)** — loads the hook at module import via `import_module_with_default`. Adds `get_client_setting_override_from_hook()` to centralize invocation + exception handling (returns `{}` on failure or when the hook is unset). Threads a new `client_setting_override` kwarg through `create_executor_from_execution` → `_get_executor_params_and_engine` → `get_client_setting_from_engine`. When the override is `None`, `get_client_setting_from_engine` invokes the hook inline; when it's a dict (including `{}`), the hook is skipped and the explicit override is merged. Plugin-supplied keys win over engine defaults.
- **[querybook/server/tasks/run_query.py](querybook/server/tasks/run_query.py)** — `run_query_task` accepts a new `client_setting_override` kwarg, normalizes a missing value to `{}` (so the worker never accidentally re-invokes the hook in a Celery context without Flask), and forwards it to `create_executor_from_execution`.
- **[querybook/server/datasources/query_execution.py](querybook/server/datasources/query_execution.py)** — `initiate_query_execution` calls `get_client_setting_override_from_hook()` while still in the Flask request that created the `QueryExecution`, and passes the result through `run_query_task.apply_async(kwargs=...)`.
- **[querybook/server/lib/query_review/base_query_review_handler.py](querybook/server/lib/query_review/base_query_review_handler.py)** — `approve_review` does the same, so peer-reviewed queries also carry the override from the approver's Flask request into the worker. Replaces the positional `args=[...]` apply_async call with a kwargs-based call for consistency.

Behavior is unchanged when no plugin implements `EXTRA_CLIENT_SETTINGS_HOOK`. Sync executor paths (`ExecuteQuery`, validators, sample queries, table upload) are unchanged at the call site — the inline branch in `get_client_setting_from_engine` handles them transparently.

### DIFF
--- a/querybook/server/lib/query_executor/executor_factory.py
+++ b/querybook/server/lib/query_executor/executor_factory.py
@@ -5,6 +5,7 @@ from const.query_execution import QueryExecutionStatus
 from lib.logger import get_logger
 from lib.query_analysis.statements import get_statement_ranges
 from lib.query_analysis.lineage import process_query
+from lib.utils.import_helper import import_module_with_default
 from logic import (
     admin as admin_logic,
     query_execution as qe_logic,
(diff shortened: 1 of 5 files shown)


querybook/server/lib/query_executor/executor_factory.py
```diff
@@ -62,13 +99,21 @@ def _get_executor_params_and_engine(
...
+) -> Dict:
     """Compute the settings passed to the query engine.
        Both engine and user must be attached to a sqlalchemy session.
 
     Args:
         engine (QueryEngine): Corresponds to the DB QueryEngine
         uid (int, optional): Optional User id executed the query. Defaults to None.
+        client_setting_override (Dict, optional): Extra kwargs merged into the
+            client_setting dict. Async callers should pass an explicit
+            override computed before queuing the Celery task. Direct/sync
```
suryabhaskarkarri, line 113: nit - Are there any uses of aync calling ?

suryabhaskarkarri (approved):

querybook/server/lib/query_executor/executor_factory.py
kgopal492: Oh it's mainly async, the query executions run asynchronously.

Sync case is only for the linter (and in case we add any more async apis in the future)

raymondtieu (approved):
lgtm

PR #1727, opened 2026-07-02
Promote Copy TSV to a standalone toolbar button

rbalicki2:
Move the "Copy to Clipboard (as TSV)" action out of the export dropdown menu and surface it as a dedicated Copy button in the statement execution bar, alongside the Export dropdown.


raymondtieu (commented):
Can you extract the `copyTSVButton` out of the dropdown component so it sits with the other sibling toolbar buttons? 
Perhaps if that effort is too high or duplicates the button handlers/dependencies, can you rename the `ResultExportDropdown` to something like `ResultExportActions` since we're always showing the copy to clipboard button now?

raymondtieu (approved):
Lgtm other than the 1 comment

PR #1728, opened 2026-07-09
Update cron to prevent invalid dates

jasoncoffman:
## Summary
The current cron validation did not ensure that the day+month combination was valid. This means a use can submit something like `2 30 2 30 *` and it will be allowed.

This PR adds an additional layer to validate that the day provided is valid


suryabhaskarkarri (approved):

wbh123456 (approved):
LGTM

PR #1731, opened 2026-08-01
feat: show messages for URL result exporters

CenturySturgeon:
## Summary
- Allow URL result exporters to return an optional structured result with `url`, `message` and `message_type`.
- Display the exporter message in the existing "export complete" modal while retaining the user clicked "View Export" action.
- Preserve compatibility with exporters that return a plain URL string.

## Test plan
- [x] `make test`
- [ ] Verified structured URL results show a warning and open the supplied URL only after clicking "View Export".
- [ ] Verified legacy plain string URL results still use the existing "export complete" modal flow.

## Compatibility
Existing URL exporters may continue returning a plain string URL unchanged. The structured result is opt-in:

```python
{
    "url": "https://example.com/export",
    "message": "Optional warning for the user",
    "message_type": "warning",
}
```


raymondtieu (approved):
lgtm

querybook/server/lib/export/base_exporter.py
```diff
@@ -12,6 +12,8 @@
 
 LOG = get_logger(__file__)
 
+ExporterResult = Union[str, dict[str, str]]
```
kgopal492: can we use a typed dict since the frontend expects a specific format

CenturySturgeon: sure thing, added typed dict!

kgopal492 (dismissed):

kgopal492 (approved):

PR #1733, opened 2026-09-10
[Querybook] Introduce PyIcebergExporter to create autogenerated Iceberg tables via PyIceberg

jennywang67:
## Summary
Introduce PyIcebergExporter to create autogenerated Iceberg tables via PyIceberg. 
PyIceberg will be the default engine when using the tool.

More details in internal PR.


querybook/server/lib/table_upload/exporter/pyiceberg_exporter.py
```diff
@@ -0,0 +1,95 @@
...
+        arrow_schema = arrow_table_temp.schema
+
+        # Handle if_exists logic
+        if_exists = self._table_config["if_exists"]
+        table_exists = False
+        try:
+            catalog.load_table(fq_table_name)
+            table_exists = True
+        except Exception:
+            pass
```
raymondtieu: Should we be catching auth/network/server errors here? `create_table` might still fail.

jennywang67: Changed behavior to raise all exceptions other than if table exists, which matches current S3ParquetExporter behavior

raymondtieu (approved):
LGTM

kgopal492 (approved):

